PR #5, opened 2016-03-01
Added docker development environment

ElijahCaine:
Run the script `dockersetup.sh -b` to build the environment.
Run the script alone to start and enter a dev container.
Run the script with the `-h` or `--help` flag for help.
Add packages to the Dockerfile when you need them installed.

Re-build when necessary.

This should help make development easier. I could be wrong.

@matthewrsj @Kennric Thoughts?

> Spoiler, it's just a bash script that wraps around some docker commands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 build
 venv
+.bash_history
+.viminfo
+.v/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+from centos:latest
+
+RUN yum -y update
+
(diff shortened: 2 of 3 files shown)


matthewrsj:
I think this is a good idea. I'll have to relearn how to use docker but that should be trivial since you already wrote the setup script. I was able to successfully run the script and generate the docker image on my machine, so at the very least it doesn't hurt anything.

tl;dr +1

Kennric:
I am going to close this and re-PR against develop, resolve conflicts and merge there.

PR #25, opened 2016-04-08
Added template for Github PR

ElijahCaine:
This PR adds a template to our pull requests so we never have to ask "What is this change and how do I test it?"

info: https://github.com/blog/2111-issue-and-pull-request-templates


Kennric:
Yes, that is a pretty cool feature, I'm up for experimenting with it.

matthewrsj:
I like it also. If it gets cumbersome or feels unnecessary we can always remove it.

Issue #28, opened 2016-04-15
Create Rufus Scheduler that runs collect_ganeti every 30 minutes [3]


matthewrsj:
Rufus needs to run the following:

1. Collectors::collect_ganeti every 30 minutes to store ganeti node information in redis.
2. All plugins::collect methods every 30 minutes (offset by 15 minutes) to move the info in redis to our DB.

matthewrsj:
Closed by #34

PR #34, opened 2016-04-18
Add rufus scheduler for collect ganeti

matthewrsj:
fixes issue #28, sort of. This just runs the collect_ganeti method because there is no plugin collect method to run yet.

## Changes in this PR.

- [X] Add rufus scheduler that runs Collectors.collect_ganeti method every 30 minutes

## Testing this PR.

1. Install redis (unless you can use docker)
2. run the `redis-server` binary in different terminal
3. run `ruby scheduler.rb` in a different terminal
4. run the `redis-cli` binary, type `keys *` --> you should get output

@osuosl/devs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ dockerfiles/app
 dockerfiles/app.env
 docker-compose.yml
 *.db
+dump.rdb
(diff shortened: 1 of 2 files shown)


Kennric:
Looks good, I think we can merge this when the p statements are removed.

matthewrsj:
@Kennric p statements removed

@ElijahCaine @alxngyn

alxngyn:
+1 for me
Any reason we don't want the P statements? Should we be logging the output somewhere?

Issue #23, opened 2016-04-05
Test infrastructure


Kennric:
create:
spec/spec_helper (sets up and runs spec tests)
add spec testing to Rakefile
add necessary gems to Gemfile
hello_world test for / (simple welcome page template)

matthewrsj:
@Kennric can this be closed?

matthewrsj:
closed by #42

PR #40, opened 2016-04-22
Kennric/database config

Kennric:
fixes issue #9

## Changes in this PR.
- [X] Creates test/dev/prod environment based settings
- [X] Connects the models to the env appropriate database
- [X] Sets up in-memory database for testing

## Testing this PR.
1. Run rake spec
2. See model tests pass

@osuosl/devs

PR #42, opened 2016-04-22
Kennric/client rspec

Kennric:
fixes issue #23

## Changes in this PR.

- [X] Add simple client routes with skeleton templates
- [X] Tests simple client routes
- [X] Adds FactoryGirl factory for clients

1. Run rake spec
2. See tests pass
3. Run ruby app.rb
4. See application run/exit without error (a run! directive needs to be added to run the dev server)

@osuosl/devs

This should be merged after #40, as it builds on that infrastructure


app.rb
```diff
@@ -1,35 +1,64 @@
...
+    erb :'clients/new'
+  end
+
+  ## TODO - move this somewhere
+  redis = Redis.new(host: ENV['REDIS_HOST'])
+
+  # TODO: Query database for each unique cluster fqdn
+  # for each cluster fqdn, append port number, endpoint, and query
+  fqdn = ['ganeti-psf.osuosl.bak', 'ganeti-civicrm.osuosl.bak']
+  fqdn.each do |name|
```
matthewrsj: Is there a reason the ganeti collector code is in app.rb?

Kennric: It was there when I branched, so I didn't touch it.

alxngyn:
+1
I agree that we should remove the collector code once we figure out how to link app.rb to the scheduler.

Gemfile
```diff
@@ -1,14 +1,39 @@
...
+  # Tools
+  gem 'dotenv'
+  gem 'thin'
+  gem 'json'
+  gem 'redis', '~>3.2'
+  gem 'rubocop', require: false
+end
+
+# Development
+group :production do
```
ElijahCaine: Should this be `group :development do`?

Rakefile
```diff
@@ -1,28 +1,24 @@
-require 'sequel'
+#require 'sequel'
+require 'rake'
+require 'rspec/core/rake_task'
+require File.expand_path '../environment.rb', __FILE__
```
ElijahCaine: I don't think it matters, but should we `require_relative` instead of this crazy `require File.expand_path ...` business? [Relevant Stack Overflow Question if `require_relative` is new to you](http://stackoverflow.com/questions/3672586/what-is-the-difference-between-require-relative-and-require-in-ruby).

app.rb
```diff
@@ -1,35 +1,64 @@
...
+require 'sinatra/base'
+require File.expand_path '../environment.rb', __FILE__
+
+# IAM - a resource usage metric collection and reporting system
+class Iam < Sinatra::Base
+  # require the models, but make sure to do it after the test db is migrated
+  require 'models'
+
+  # load up all the plugins
+  Dir[File.dirname(__FILE__) + '/plugins/*.rb'].each { |file| require file }
```
ElijahCaine: Are plugins single files or did we decide they were in their own directory?

app.rb
ElijahCaine: I agree that we should get rid of anything in `app.rb` that isn't necessary. It should just have enough to run the application for testing purposes.

app.rb
matthewrsj: They are supposed to be in `plugins/<plugin_name>` so this should be changed to reflect that... somehow.

Rakefile
Kennric: Probably, but that was outside the scope of the pr, this is just trying to get an rspec framework into develop

app.rb
Kennric: we did make that change, but that wasn't in develop when I branched - I want to keep the scope of this PR pretty narrow, but since this also will relate to how we load plugin rspec, I'll see if I can fix it.

Kennric:
Request for comment: @matthewrsj @ElijahCaine @alxngyn 

Plugin file naming scheme - for easy of loading plugin files, we want predictably distinct file names for the plugin, spec and migration files under plugins/plugin_name/

The current app code loads all .rb files - obviously a problem when we add spec and migration files. We could exclude known non-plugin-code files, like *_spec.rb and *_migration.rb, but I propose just naming the files with an explicit pattern, so we can find spec, code and migrations easily without danger of pulling in any other file that might exist - some variation of:

```
/plugins/<name>/
    <name>_spec.rb
    <name>_plugin.rb
    <name>_migration.rb
    some_other_file_for_some_reason.rb
    readme.txt
```
or: 
```
/plugins/<name>/
    spec.rb
    plugin.rb
    migration.rb
    some_other_file_for_some_reason.rb
    readme.txt
```

matthewrsj:
@Kennric that makes sense to me, seems like an easy way to do it. I like your second example, but only because it requires less `*`s.

Rakefile
ElijahCaine: Fair enough. Just commenting on the green lines.

ElijahCaine:
@Kennric requested comment: I prefer option 2 (`<name>/<thing>.rb`) since the files are already namespaced by their directory we don't need to double-namespace the files with `<name>/<name>_<thing>.rb`.

alxngyn:
@Kennric I'm fine with either as long as we throw in a README.md into the plugins/ documenting it.

alxngyn:
So after working on creating the plugin migrations file, Sequel migration needs that specific format of `#_name.rb` to keep track of migration version and the name is basically a keyword of the change. 

I change my vote for a migrations folder so we have `plugins/<plugin name>/migrations/###_change.rb` to keep migrations organized for each plugin. For iterating through plugin migrations, I believe we can just change it to `plugins/**/migrations/*.rb`

matthewrsj:
+1, we can make any changes later.

Issue #31, opened 2016-04-15
Create Report Method for disk_sizes Plugin [8]


matthewrsj:
The `report` method performs the following tasks when called by the `/report` endpoint:

1. Pull `disk_sizes` information for the specified node out of our database. `report` should pull data from a provided `start_date` to `end_date`.
2. Perform business logic (to be defined) on the data. This may be an average or something else.
3. Return information to the `/report` endpoint and output as csv file.

matthewrsj:
Fixed by #43

PR #35, opened 2016-04-18
Add store method for disk_sizes plugin

matthewrsj:
Fixes #30 
Fixes #32 
Fixes #33 

## Changes in this PR.

- [x] Create store method for disk_sizes plugin
- [x] Schedule the disk_sizes plugin store method
- [x] Tests the store method
- [x] Creates the register method
- [x] Creates the migrations file

## Testing this PR.

1. `docker-compose run dev bash` after making sure your app.env is correct
2. `rake spec`

***OR*** (a more hands-on method)

1. `docker-compose run dev bash` after making sure your app.env is correct
2. On line 15 of scheduler.rb change `'15m'` to `4`. Note you need to remove the quotes. You can also `puts` something after the `.each` loop if you want to see when it is finished.
4. `rake migrate && ruby scheduler.rb` then wait 15-20 seconds before pressing CTRL-C (it does it's work silently and immediately, then every 30 minutes). If you added a `puts` statement above you will see it when this finishes.
7. Open sqlite3 database, `sqlite3 /tmp/development.sqlite`
8. `select * from disk_size_measurements;`

You should see something like this:
```
1||2016-05-06 22:17:15.453218|mandrivausers2.osuosl.org|12800|1
2||2016-05-06 22:17:15.631977|deluge.osuosl.org|20480|1
3||2016-05-06 22:17:15.817240|rtvstaging2-web.osuosl.org|10752|1
4||2016-05-06 22:17:15.995706|efs.osuosl.org|11264|1
5||2016-05-06 22:17:16.166123|fossology.osuosl.org|15360|1
6||2016-05-06 22:17:16.349770|pgd-app1.osuosl.org|15360|1
7||2016-05-06 22:17:16.533202|systers-prod.osuosl.org|20480|1
8||2016-05-06 22:17:16.717963|jaws2.osuosl.org|16384|1
9||2016-05-06 22:17:16.903161|oprod-controller.osuosl.org|10240|1
10||2016-05-06 22:17:17.059477|teadmin-r.osuosl.org|5120|1
11||2016-05-06 22:17:17.217986|spacewalk.osuosl.org|40960|1
12||2016-05-06 22:17:17.629046|rtvprodvr-util.osuosl.org|38912|1
13||2016-05-06 22:17:17.793969|connectopensource.osuosl.org|20480|1
14||2016-05-06 22:17:17.952897|crosstool-ng.osuosl.org|10752|1
...
```

I get 187 lines.

@osuosl/devs

Issue #30, opened 2016-04-15
Create Store Method for disk_sizes Plugin [5]


matthewrsj:
The `collect` method performs the following tasks when called by the rufus scheduler:

1. Pull `disk_sizes` information for the specified node out of redis.
2. Sum all `disk_sizes` values (`disk_sizes` is a list of integers).
3. Store `sum` in the `disk_sizes_measurements` table in the database.

PR #43, opened 2016-04-26
Alxngyn/disk size report

alxngyn:
### Updated PR @ [link](#issuecomment-217589874)


Kennric:
This looks pretty good so far, but I'd like to see the following changes:

move the plugin code and spec both into plugins/disk_size/
edit Rakefile to also find spec files in plugins/
add something to the tests that actually confirm that the report method exists (current test doesn't actually do that)

alxngyn:
- [x] move the plugin code and spec both into plugins/disk_size/
- [x] edit Rakefile to also find spec files in plugins/
- [x] add something to the tests that actually confirm that the report method exists (current test doesn't actually do that) 
 - `plugins/disk_size/disk-size_spec.rb` unit tests if the 3 main methods exist

### Testing this PR (update)

1. Run `rake spec`
2. Should see two runs of Rspec, one for the main app and one for the plugin, I couldn't find any doc or SO post about running RSpec in two different directories in one run

plugins/disk_size/disk_sizes.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    # puts 'node: ' + node
+    # puts 'start: ' + start_time.to_s
+    # puts 'end: ' + end_time.to_s
+
+    # gather data from Db here
+    # dataset = DB[:disk_size_measurement]
+    redis = Redis.new(host: ENV['REDIS_HOST'])
+    data_hash = {}
+    servers = redis.keys
```
matthewrsj: I don't think we want to get all keys (even when we are using a database), unless specifically requested by the user.

matthewrsj:
You could simplify this method considerably by either getting the single node asked for or all as a default. You could do something like this:

```ruby
def report(node='*')
```

Then if they just ask for `DiskSizePlugin.report()` they get all nodes, but if they ask for `DiskSizePlugin.report(fqdn='nodefqdn') they get just what they are asking for.

In the method you would either query the db for the one node or loop through all. You could even put a single node in a list so you can use the same logic for both. As it is you aren't using the `node` argument at all.

I'm not convinced we should merge code that uses logic we don't want in the finished product. This communicates with redis when it should be communicating with a DB. Basically, this method does the combined work of `collect` and `report` by bypassing the database, which we don't want to do.

I think we should wait until we get postgres figured out and start using it before finishing this work since so much of it will have to change.

alxngyn:
Thanks for the feedback!
I'll wait for/work on the `collect` method then I'll reset this branch with changes for `report` with DB integration, node parameters, etc.

matthewrsj:
@alxngyn great. I have a collect method in #30 that I've been working on. I plan to work on making sure we can get the DB up and running later after some higher priority work I have to do. Feel free to work on the DB stuff yourself too.

alxngyn:
fixes issue #31

## Changes in this PR.


- [X] Created disk_size report method
- [X] Created test_data.rb to fill database with dummy data

## Testing this PR.


To test the report method (yeah it could be better):
1. Run docker `$ docker-compose run dev bash`
2. Migrate app DB `$ rake migrate`
3. Uncomment the last 3 lines at the bottom of plugin.rb `DiskSize.new.register...`
4. Run plugin migrate to DB `$ ruby plugins/disk_sizes/plugin.rb`
5. Run test_data to fill DB with data, `$ ruby plugin/disk_sizes/test_data.rb`
6. Run plugin.rb again and there should be a 2 JSON outputs from the DB. `$ ruby plugins/disk_sizes/plugin.rb`

Expected Output
```
[{"id":22,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1234},{"id":23,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1234},{"id":24,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1235},{"id":25,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1234},{"id":26,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1235},{"id":27,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1235},{"id":28,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1239},{"id":29,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1236},{"id":30,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1237},{"id":31,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1242},{"id":32,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1234}]
[{"id":1,"node_resource":null,"time":"2016-05-06 23:55:03 +0000","node":"FACEYMYBOOKY.com0","value":1234},{"id":2,"node_resource":null,"time":"2016-05-06 23:55:03 +0000","node":"FACEYMYBOOKY.com1","value":1234},{"id":3,"node_resource":null,"time":"2016-05-06 23:55:03 +0000","node":"FACEYMYBOOKY.com2","value":1234},{"id":4,"node_resource":null,"time":"2016-05-06 23:55:03 +0000","node":"FACEYMYBOOKY.com3","value":1235},{"id":5,"node_resource":null,"time":"2016-05-06 23:55:03 +0000","node":"FACEYMYBOOKY.com4","value":1234},{"id":6,"node_resource":null,"time":"2016-05-06 23:55:04 +0000","node":"FACEYMYBOOKY.com5","value":1235},{"id":7,"node_resource":null,"time":"2016-05-06 23:55:04 +0000","node":"FACEYMYBOOKY.com6","value":1238},{"id":8,"node_resource":null,"time":"2016-05-06 23:55:04 +0000","node":"FACEYMYBOOKY.com7","value":1238},{"id":9,"node_resource":null,"time":"2016-05-06 23:55:04 +0000","node":"FACEYMYBOOKY.com8","value":1237},{"id":10,"node_resource":null,"time":"2016-05-06 23:55:04 +0000","node":"FACEYMYBOOKY.com9","value":1242},{"id":11,"node_resource":null,"time":"2016-05-06 23:55:04 +0000","node":"FACEYMYBOOKY.com10","value":1239},{"id":12,"node_resource":null,"time":"2016-05-06 23:55:04 +0000","node":"FACEYMYBOOKY.com11","value":1238},{"id":13,"node_resource":null,"time":"2016-05-06 23:55:05 +0000","node":"FACEYMYBOOKY.com12","value":1237},{"id":14,"node_resource":null,"time":"2016-05-06 23:55:05 +0000","node":"FACEYMYBOOKY.com13","value":1241},{"id":15,"node_resource":null,"time":"2016-05-06 23:55:05 +0000","node":"FACEYMYBOOKY.com14","value":1238},{"id":16,"node_resource":null,"time":"2016-05-06 23:55:05 +0000","node":"FACEYMYBOOKY.com15","value":1238},{"id":17,"node_resource":null,"time":"2016-05-06 23:55:05 +0000","node":"FACEYMYBOOKY.com16","value":1237},{"id":18,"node_resource":null,"time":"2016-05-06 23:55:05 +0000","node":"FACEYMYBOOKY.com17","value":1247},{"id":19,"node_resource":null,"time":"2016-05-06 23:55:05 +0000","node":"FACEYMYBOOKY.com18","value":1251},{"id":20,"node_resource":null,"time":"2016-05-06 23:55:05 +0000","node":"FACEYMYBOOKY.com19","value":1251},{"id":21,"node_resource":null,"time":"2016-05-06 23:55:06 +0000","node":"FACEYMYBOOKY.com20","value":1247},{"id":22,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1234},{"id":23,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1234},{"id":24,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1235},{"id":25,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1234},{"id":26,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1235},{"id":27,"node_resource":null,"time":"2016-05-06 23:55:08 +0000","node":"FACEYMYBOOKY.com","value":1235},{"id":28,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1239},{"id":29,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1236},{"id":30,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1237},{"id":31,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1242},{"id":32,"node_resource":null,"time":"2016-05-06 23:55:09 +0000","node":"FACEYMYBOOKY.com","value":1234}]

```


@osuosl/devs

ElijahCaine:
LGTM +1
Pretty well commented too. Thanks for that :smile:

matthewrsj:
@alxngyn can you add spec tests first? Also, can you change the naming scheme to what was introduced in #35 and rebase?

alxngyn:
Fixed a rubocop complaint and rebased with develope

alxngyn:
added spec tests. Since report is dependent of the data in the DB and a time range, some tests may fail (stale data) but shouldn't once we have a running app.

matthewrsj:
We should think of a way to dynamically test the logic instead of relying on existing database content. Can we insert some test data to test on, then remove it after the tests run?

alxngyn:
@matthewrsj You're right. I can make that happen with some Sequel inserts and deletes so the tests are independent from everything.

plugins/disk_sizes/plugin.rb
```diff
@@ -10,10 +10,11 @@ def register
                           storage_table: 'disk_size_measurements',
                           units: 'bytes')
     # execute migration
-    Sequel::Migrator.run(Iam.settings.DB,
-                         '/data/code/plugins/disk_sizes/migrations',
-                         column: :disk_size_ver
-                        )
+    Sequel::Migrator.run(
+                        Iam.settings.DB,
+                        '/data/code/plugins/disk_sizes/migrations',
```
ElijahCaine: This path should probably be relative to the base of the repo, not absolute.

alxngyn:
updated rspecs with DB data and updated PR with `rake spec` how-to

matthewrsj:
@alxngyn do you still need the test data script?

alxngyn:
@matthewrsj you're right. I'll delete the file

matthewrsj:
+1

alxngyn:
Does @ElijahCaine's +1 still count in the OSL even after those changes?

matthewrsj:
@alxngyn nope, you need two +1s after the very last change, even if the last change is just a rebase.

ElijahCaine:
Tests look good. I do think that path should be generated at runtime and not absolute. It makes too many assumptions about our environment. Pending that feedback being addressed I'll keep my +1 on this.

alxngyn:
@ElijahCaine, the future error solver, is right.
Made migration path relative and fixed rubocop complaint

plugins/disk_sizes/plugin.rb
```diff
@@ -10,11 +10,10 @@ def register
...
                           units: 'bytes')
     # execute migration
-    Sequel::Migrator.run(
-                        Iam.settings.DB,
-                        '/data/code/plugins/disk_sizes/migrations',
-                        column: :disk_size_ver
-    )
+    path = File.dirname(__FILE__) + '/migrations'
+    Sequel::Migrator.run(Iam.settings.DB,
+                         path,
```
matthewrsj: Instead of creating a variable you only use once, can you put `File.dirname(__FILE__) + '/migrations'` here directly?

plugins/disk_sizes/plugin.rb
```diff
@@ -10,10 +10,10 @@ def register
                           storage_table: 'disk_size_measurements',
                           units: 'bytes')
     # execute migration
+    path = File.dirname(__FILE__) + '/migrations'
     Sequel::Migrator.run(Iam.settings.DB,
-                         '/data/code/plugins/disk_sizes/migrations',
-                         column: :disk_size_ver
-                        )
+                         path,
```
matthewrsj: Instead of creating a variable you only use once, can you put `File.dirname(__FILE__) + '/migrations'` here directly?

ElijahCaine: @matthewrsj I would rather have an expressive variable that gets used once than nonsense.

alxngyn: removed the one time use var

matthewrsj: I disagree that `File...` is unintuitive. From what I've seen it is pretty widely used in Ruby.

matthewrsj:
Well, this gets a +1 from me, @ElijahCaine may disagree.

alxngyn:
x8fdaj8

ElijahCaine:
I'm going to keep my +1 on this, but only because it's a blocker and I loathe nit-picking.

Issue #60, opened 2016-05-16
Command rake spec should run in one spec run, not two


matthewrsj:
When `rake spec` is run, it should complete in one rspec run. Fixed by changing `Dir.glob` pattern to include all relevant directories.

matthewrsj:
Fixed by #61

PR #61, opened 2016-05-16
Rake spec should only use one rspec run

matthewrsj:
fixes issue #60 

## Changes in this PR.

- [X] `rake spec` runs all spec tests in one run

## Testing this PR.

2. `rake spec`

### Expected Output.

```
[root@c499ce58a036 code]# rake spec
/usr/local/bin/ruby -I/usr/local/lib/ruby/gems/2.3.0/gems/rspec-support-3.4.1/lib:/usr/local/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/lib /usr/local/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/exe/rspec plugins/disk_sizes/spec.rb spec/models/client_spec.rb spec/routes/app_spec.rb spec/routes/client_spec.rb --format documentation

DiskSize
  .register does not raise an error when invoked
  .register actually actually creates a disk_size_measurements table
  .store does not fail with valid data
  .store fails when not passed node name
undefined method `tr' for nil:NilClass
  .store does not crash when passed improperly formatted data
  .store properly sums all disk sizes when storing in DB
  .report method
    returns empty on unknown node
    should return empty on invalid day input
    should return empty on invalid fqdn input

The Client Model and table
  initially has no clients
  has a model name
  can create a client
  cannot create two clients with the same name
  can create two clients with different names
  cannot create a client without required fields
  cannot create a client without required fields
  can delete a client
  can update a client
  cannot update a client to have a non-unique name

The IAM Application
  says hello

The Client list endpoint
"Client X"
  responds OK
"Client X"
  contains the name of a client we added

Finished in 0.07318 seconds (files took 0.31841 seconds to load)
24 examples, 0 failures
```

@osuosl/devs

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -34,18 +34,9 @@ task :rubocop do
 end
 
 # rake spec
-task spec: [:specs, :spec_plugins] do
-end
-
-# spec files in the spec/ folder
-RSpec::Core::RakeTask.new(:specs) do |t|
+RSpec::Core::RakeTask.new(:spec) do |t|
   ENV['RACK_ENV'] = 'test'
-  t.pattern = Dir.glob('spec/**/*_spec.rb')
+  t.pattern = Dir.glob('{spec,plugins}/**/*spec.rb')
   t.rspec_opts = '--format documentation'
 end
 
-# spec files in the plugins folders
-RSpec::Core::RakeTask.new(:spec_plugins) do |t|
-  t.pattern = Dir.glob('plugins/**/*spec.rb')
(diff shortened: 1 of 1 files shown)


ElijahCaine:
*This pleases me* +1

matthewrsj:
This is also simple enough to merit one +1. Merging...

alxngyn:
+1
Those brackets were what stumped me when I tried to do it.

Issue #48, opened 2016-05-06
vcpu count plugin


Kennric:

ElijahCaine:
Questions to answer before writing the plugin:
[ ] What is the unit of measurement for a VCPU?
[ ] Which if our API's provide this information and in what form?

matthewrsj:
I would like to work on this

matthewrsj:
Looks like the only available fields regarding vcpus are the count of vcpus available:

```
...
"oper_vcpus"=>8,
...
"custom_beparams"=>{"minmem"=>2048,
                    "vcpus"=>8,
                    "maxmem"=>2048},
...
"beparams"=>{"auto_balance"=>true,
             "vcpus"=>8,
             ...
...
```

ElijahCaine:
@matthewrsj In that case the unit of `vcpu` (or I guess `cpu` if we're also working with physical hardware?) is most appropriate.

matthewrsj:
@ElijahCaine I see the light. Type changed to `vcpu`

matthewrsj:
Resolved by #62

PR #62, opened 2016-05-16
VCPUCount Plugin

matthewrsj:
fixes issue #48 

## Changes in this PR.

- [X] Adds plugin methods for VCPUCount plugin (`.register`, `.store`, and `.report`)
- [X] Adds migration file for VCPUCount plugin
- [X] Adds spec tests for VCPUCount plugin

## Testing this PR.

1. `docker-comopse run dev bash` after making sure `dockerfiles/app.env` is correct
2. `rake spec`

### Expected Output.

```
[root@ff066777bf17 code]# rake spec
[... passing test output ...]

VCPUCount plugin
  .register method
    creates a vcpu_count_measurements table
  .store method
    does not fail with valid data
  .report method
    returns empty on unknown node

[... passing test output ...]

Finished in 0.05974 seconds (files took 0.37272 seconds to load)
[number] examples, 0 failures
```

@osuosl/devs


plugins/vcpu_count/spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      expect { VCPUCount.new.register }.to_not raise_error
+    end
+
+    it 'creates a vcpu_count_measurements table' do
+      # Table shouldn't exist before registration
+      expect do
+        Iam.settings.DB.table_exists?(:vcpu_count_measurements).to be_false
+      end
+
+      VCPUCount.new.register
```
ElijahCaine: Is there a reason we have to explicitly call the `register` method on the plugin? I recall we talked about calling the register method by default when a Plug-in object is created.

ElijahCaine:
Just a nit-pick comment. Pending that being addressed +1 from me.

plugins/vcpu_count/spec.rb
matthewrsj: @ElijahCaine That does make more sense, since everywhere we use the plugin we register it first to make sure the migrations are called. Will fix.

datastruct.erb
```diff
@@ -1 +1,7 @@
-{"disk_sizes": "<%= disk_sizes_meas %>", "disk_usage": "<%= disk_usage_meas %>", "disk_template": "<%= disk_template_meas %>", "num_cpus": "<%= num_cpus_meas %>", "total_ram": "<%= total_ram_meas %>", "active": <%= active_meas %> }
+{"disk_sizes": "<%= disk_sizes_meas %>",
+ "disk_usage": "<%= disk_usage_meas %>",
+ "disk_template": "<%= disk_template_meas %>",
+ "num_cpus": "<%= num_cpus_meas %>",
+ "total_ram": "<%= total_ram_meas %>",
+ "active": <%= active_meas %>
+ }
```
alxngyn: File leak?
I don't think this template belongs in this PR

plugins/VCPUCount/plugin.rb
```diff
@@ -0,0 +1,61 @@
...
+    # else return data filtered with fqdn name
+    else
+      dataset = @database[@table].where(node: fqdn)
+                                 .where(created: start_time..end_time)
+    end
+    # format and make json/csv thing
+    dataset.all.to_json
+  end
+
+  VCPUCount.new.register
```
alxngyn: Register method should be moved to initialized like the fix in #66

alxngyn:
+1 once those two, somewhat minor, fixes goes in

datastruct.erb
matthewrsj: I did that to make it easier to read. If you want I can move it to a separate PR but since it was so simple I left it in here.

alxngyn: Misunderstood, never mind my feedback on this. 
readability change +1

alxngyn:
good work +1

Issue #65, opened 2016-05-16
Plugin name in table != plugin folder name or actual plugin name


matthewrsj:
for example:
```
$ ls plugins/
disk_sizes
```

but in `plugins/disk_sizes/plugin.rb`:
```
# Disk Sizes data plugin
class DiskSize
  def register
    Plugin.find_or_create(name: 'disk_size', # create entry in Plugins table
    ...
```

This is an issue when telling our scheduler to call each plugin based on name in the plugin table.

alxngyn:
resolved by #66

PR #66, opened 2016-05-16
Change folder name and table entry name to match plugin name

matthewrsj:
fixes issue #65 

## Changes in this PR.

- [X] Changes folder and plugin table entry to match plugin name
- [X] Call register method on plugin creation

## Testing this PR.

2. `rake spec`

**Also**

1. Visually inspect file. We do not *yet* have code that this will affect (this was a preemptive bugfix)

### Expected Output.

```
$ rake spec
[... passing tests ...]
```

You should also see that the `disk_sizes` folder was renamed to `DiskSize` to match the plugin name.

@osuosl/devs


plugins/DiskSize/plugin.rb
```diff
@@ -63,9 +63,7 @@ def report(fqdn = '*', days = 1)
     # format and make json/csv thing
     dataset.all.to_json
   end
+
+  DiskSize.new.register
```
alxngyn: I think the register method should be called initialize if we want it be called on first run or just removed completely

matthewrsj: absolutely. You are correct about the initialized method.

alxngyn:
+1

PR #68, opened 2016-05-16
Update Scheduler Store Task

### LINKED ISSUE #67: The scheduler should dynamically pull in plugins and run them

matthewrsj:
fixes #67 

#### Rakefile:
- [X] Create `plugins` task to register plugins with `rake`
  - [X] Load/require plugins by name from plugin folder names
  - [X] Register each plugin

#### Scheduler:
- [X] Call `rake plugins`
- [X] Load/require plugins by name from plugins table
- [X] Call store method on each plugin

## Testing this PR.

2. `rake spec` - Not really necessary for these changes, but just make sure nothing fails.
3. `rake migrate`
4. In `scheduler.rb`: change `'15m'` on line 16 to `4` (note, no quotes `'`) and add `puts 'done'` before the last `end` on line 29.
5. `ruby scheduler.rb` and wait for `done` to print, then press CTRL-C
6. `sqlite3 /tmp/development.sqlite`
7. `select * from disk_size_measurements;`
8. `select * from vcpu_count_measurements;`

### Expected Output.

```
$ rake spec
[... passing tests ...]
$ rake migrate
Migrating to latest
$ ruby scheduler.rb
[... a lot of time passing ...]
done
$ sqlite3 /tmp/development.sqlite
SQLite version 3.7.17 2013-05-20 00:56:22
Enter ".help" for instructions
Enter SQL statements terminated with a ";"
sqlite> select * from disk_size_measurements;
[... 180+ rows of disk size measurements ...]
sqlite> select * from vcpu_count_measurements;
[... 180+ rows of vcpu count measurements ...]
```
@osuosl/devs

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -42,6 +42,7 @@ end
 
 # rake plugins
 task :plugins do
+  puts 'Registering plugins'
   # Get name of each plugin from folder name.
   # At this point, we are assuming the plugin is *not* registered and is
   # therefore not in the plugins table.
--- a/Rakefile
+++ b/Rakefile
@@ -50,9 +50,7 @@ task :plugins do
     # The actual name is the last part of the filename (plugin/<name>).
     plugin_name = File.split(name)[-1]
     # Require each plugin and register.
-    Dir["#{name}/plugin.rb"].each do |plugin|
-      require plugin
-      Object.const_get(plugin_name).new
-    end
+    require "#{name}/plugin.rb"
+    Object.const_get(plugin_name).new
   end
 end
--- a/Rakefile
+++ b/Rakefile
@@ -40,3 +40,18 @@ RSpec::Core::RakeTask.new(:spec) do |t|
   t.rspec_opts = '--format documentation'
 end
 
+# rake plugins
+task :plugins do
+  # Get name of each plugin from folder name.
+  # At this point, we are assuming the plugin is *not* registered and is
+  # therefore not in the plugins table.
+  Dir['plugins/*'].each do |name|
+    # The actual name is the last part of the filename (plugin/<name>).
+    plugin_name = File.split(name)[-1]
+    # Require each plugin and register.
+    Dir["#{name}/plugin.rb"].each do |plugin|
+      require plugin
+      Object.const_get(plugin_name).new
+    end
+  end
+end
--- a/Rakefile
+++ b/Rakefile
@@ -46,9 +46,9 @@ task :plugins do
   # Get name of each plugin from folder name. Require each plugin and register.
   # At this point, we are assuming the plugin is *not* registered and is
   # therefore not in the plugins table.
-  Dir['plugins/*'].each do |name|
-    require "#{name}/plugin.rb"
-    # The actual name is the last part of the filename (plugin/<name>).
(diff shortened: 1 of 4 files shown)


alxngyn:
The plugins will register via app.rb right?
I'm just thinking about the first run and how the DB will be empty when scheduler looks for plugins

matthewrsj:
Yeah, this still needs work

matthewrsj:
@alxngyn Actually, I think it makes more sense to make it a rake task, which I have done here. Thoughts?

Rakefile
```diff
@@ -40,3 +40,15 @@ RSpec::Core::RakeTask.new(:spec) do |t|
   t.rspec_opts = '--format documentation'
 end
 
+# rake plugins
+task :plugins do
+  puts 'Registering plugins'
+  # Get name of each plugin from folder name. Require each plugin and register.
+  # At this point, we are assuming the plugin is *not* registered and is
+  # therefore not in the plugins table.
+  Dir['plugins/*'].each do |name|
```
ElijahCaine: Why isn't this just `Dir[plugins/*/plugin.rb].each`?

matthewrsj: Because before I do that I need to capture the name of the plugin that lives at that /*/ level. But I just realized I can split and get the [-2] or [1] position for the name...

will fix.

ElijahCaine:
One nit-pick. Everything else looks good though. Pending +1.

alxngyn:
@matthewrsj I don't think the rake file should be messing with our app.rb aside from starting and stopping it. We can make the scheduler check for any new plugins and run their register method 5 minutes before the store method. Therefore we don't have to do the extra step of running a `rake migration` after each install of a plugin.  
Also, after writing this, we will need a way to toggle/disable plugins

Rakefile
matthewrsj: @ElijahCaine actually, this is the best way to do this because `File.split()` only splits off the last part of the path. If I use `Dir[plugins/*/plugin.rb].each` the path will be `plugins/DiskSize/plugin.rb` for the DiskSize plugin. When I split that I get an array `['plugins/DiskSize', 'plugin.rb']`. I would then need to split the array's zeroeth element again to get the `DiskSize` string.

Basically, using `Dir[plugins/*/plugin.rb].each` means more complexity than using `Dir['plugins/*'].each` and `require "#{name}/plugin.rb"`.

ElijahCaine: Okay. I did some hacking and this is a fine solution. The only way I found that I like a little more was:
```Ruby
  Dir['plugins/*/plugin.rb'].each do |name|
    require name
    # The actual name is the last part of the filename (plugin/<name>).
    Object.const_get(name.split('/')[1]).new
  end
```
But it's not difficult to understand what is happening here so I'm okay leaving it.

matthewrsj:
@ElijahCaine Changed rake task syntax to what you suggested. I also liked it slightly better.

@alxngyn I kept the rake task and added a call to it in the scheduler, as was talked about in 5/20 sprint meeting.

alxngyn:
+1 

Took me forever to realize that `rake plugins` in scheduler is a command and not some string

Kennric:
+1

PR #73, opened 2016-05-20
Log detailed errors in plugin store methods

### LINKED ISSUE #72: Log more error information in store methods
In our current plugins' store methods, we don't adequately handle empty or `'unknown'` results from redis.

In VCPUCount we get a `no implicit conversion of String into Integer` message on the value assignment when `node_info['num_cpus']` is set to 'unknown' by the collector. We should capture more information and log it with `STDERR.puts`.
```ruby
  def store(fqdn)
      ...
      value:         node_info['num_cpus'].to_i,
      ...
  rescue => e        # Don't crash on errors
    STDERR.puts e    # Log the error
  end
```

Similarly, we get a `undefined method 'tr' for nil:NilClass` in the DiskSizes plugin on the injection:
```ruby
  def store(fqdn)
      ...
      node:          fqdn,
      value:         node_info['disk_sizes']
                       .tr('[]', '')    # remove '[' and ']' from string
                       .split(',')      # split into array of strings on ','
                       .map(&:to_i)     # convert each element to integer
                       .inject(0, :+),  # inject a + method to sum the array
      ...
  rescue => e        # Don't crash on errors
    STDERR.puts e    # Log the error
  end
```

matthewrsj:
fixes #72 

## Changes in this PR.

- [X] Add error checking to `DiskSizes.store`
- [X] Add error checking to `VCPUCount.store`
- [X] Fix minor rubocop issue with VCPUCount spec

## Testing this PR.

2. `rake spec`

### Expected Output.

```
$ rake spec
DiskSize plugin
  .register method
    creates a disk_size_measurements table
  .store method
    does not fail with valid data
No disk_sizes information for badnode: {"disk_size"=>"[10,20]", "active"=>true}
    properly sums all disk sizes when storing in DB
  .report method
    returns empty on unknown node
    should return empty on invalid day input
    should return empty on invalid fqdn input

VCPUCount plugin
  .register method
    creates a vcpu_count_measurements table
  .store method
    does not fail with valid data
No num_cpus information for badnode: {"num_cpu"=>"eight", "active"=>true}
  .report method
    returns empty on unknown node
[... more passing tests ...]
```

@osuosl/devs

### DIFF
--- a/plugins/DiskSize/plugin.rb
+++ b/plugins/DiskSize/plugin.rb
@@ -26,6 +26,11 @@ class DiskSize
     # Pull node information from redis as a ruby hash
     node_info = JSON.parse(@redis.get(fqdn))
 
+    # Check for valid data
+    if node_info['disk_sizes'].nil? || node_info['disk_sizes'] == 'unknown'
+      raise "No disk_sizes information for #{fqdn}"
+    end
+
     # Insert data into disk_size_measurements table
     @database[@table].insert(
       node:          fqdn,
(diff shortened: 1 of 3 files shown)


alxngyn:
+1

We're going to need to implement a base class for plugins so we don't have to do changes like this in the future.

ElijahCaine:
The output matches the expected output. +1

PR #82, opened 2016-05-24
Add DiskTemplate plugin

### LINKED ISSUE #74: DiskTemplate plugin
Records if it is DRBD or not (0 or 1 or t/f)

matthewrsj:
fixes #74 

## Changes in this PR.

- [X] Add DiskTemplate `plugin.rb`
- [X] Add DiskTemplate migrations
- [X] Add DiskTemplate spec tests

## Testing this PR.

2. `rake spec`

### Expected Output.

```
$ rake spec
[... passing tests ...]

DiskTemplate plugin
  .register method
    creates a disk_template_measurements table
  .store method
    does not fail with valid data
No disk_template information for badnode: {"disk_type"=>"1", "active"=>true}
  .report method
    returns empty on unknown node

[... passing tests ...]
```

@osuosl/devs

### DIFF
--- a/plugins/DiskTemplate/migrations/001_migrations.rb
+++ b/plugins/DiskTemplate/migrations/001_migrations.rb
@@ -1,0 +1,15 @@
+# Disk_Size's migration file
+
+Sequel.migration do
+  transaction
+  change do
+    create_table(:disk_template_measurements) do # Plugin's DB model
+      primary_key   :id
+      foreign_key   :node_resource
+      Time          :created, null: false
+      String        :node
+      String        :value
+      Boolean       :active
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


alxngyn:
+1 
I think the name should be diskType instead of diskTemplate but it doesn't matter

matthewrsj:
@alxngyn I went with disk template because that is what ganeti calls it. I believe `template` is more descriptive than `type` anyways.

ElijahCaine:
Output matches that listed in the PR.
Tests look solid. Implementation looks good too.
+1 from me.

ElijahCaine:
Output matches that listed in the PR.
Tests look solid. Implementation looks good too.
+1 from me.

PR #84, opened 2016-05-27
fixed --service-ports docker command

alxngyn:
## Changes in this PR.

- [X] Fixed README.md so that the proper docker command is used when running `ruby app.rb`

## Testing this PR.

1. Look at changes
2. Run it if you want to use the app with the forwarded ports


ElijahCaine:
Good catch. +1

Issue #22, opened 2016-04-05
Node Resource CRUD


Kennric:
Basic CRUD endpoints for Node Resources:

get /nodes -> list of all nodes
get /node -> node creation form
get /node/:id -> node edit form
put /node -> create new node
post /node/:id -> update existing node

Kennric:
Notes:
Node form should allow selecting pre-existing Project
name should validate for fqdn
type should be a set list of available types

PR #85, opened 2016-05-27
Node resource CRUD

alxngyn:
fixes issue #22 

## Changes in this PR.

- [X] Added spec tests for `/node` endpoint
- [X] Added `/node` endpoint similar to project and client in app.rb

## Testing this PR.

1. Start container
2. Run rake spec and see tests pass
3. Side note: going to `localhost:4567/nodes` will result in a blank page since we did not run the scheduler to fill the DB with data

### Expected Output.

```
$ rake spec
[ should pass ]

### CODE REVIEW

**Comment by alxngyn** [comment]
one `rake spec` test will fail without name validation
Blocked by #86

**Comment by alxngyn** [comment]
rebased with develop, tests should now all pass

File: routes/nodes.rb [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+        @nodes = NodeResource.all
+        erb :'nodes/index'
+      end
+
+      # This could also be PUT
+      app.post '/nodes' do
+        # recieve new node
+        node = NodeResource.create(project_id: Iam.settings.DB[:projects]
+                                                .where(name: params[:project_name])
+                                                .get(:id) || node.project_id,
```
ElijahCaine: Does the `|| node.project_id` work if we're creating the `node` being referred to?

routes/nodes.rb
```diff
@@ -0,0 +1,74 @@
...
+        # recieve an updated node
+        node = NodeResource[id: params[:id]]
+
+        node.update(project_id: Iam.settings.DB[:projects]
+                                  .where(name: params[:project_name])
+                                  .get(:id) || node.project_id,
+                    name:       params[:name] || node.name,
+                    type:       params[:type] || node.type,
+                    cluster:    params[:cluster] || node.cluster,
+                    created:    params[:created] || node.created,
```
ElijahCaine: Do we actually want people to be able to edit `created` and `updated` times?

ElijahCaine:
This looks pretty good. The tests seem to cover the basics and the implementation is good. The tests seem to be covering the NodeResource model, do we have any interest in testing the endpoints?

For instance we sould use the endpoint and checking the database to see if the right thing happened? I don't think we've been doing this but I think we really should.

Thoughts @Kennric @alxngyn @athai @subnomo

routes/nodes.rb
alxngyn: Oops, that won't work.
It should be OR'd with a blank('') if no project is found

routes/nodes.rb
alxngyn: You're right. Good catch, I'll take those two out.

alxngyn:
Added ~9 new tests for the /node endpoints. 
`rake spec` should all pass

ElijahCaine:
@alxngyn After merging the changes from #87 I am able to run the app and it works seamlessly, so when we merge things should work themselves out. If you want we can rebase with develop before merging just to be sure, I don't care either way. The tests pass and seem thorough.
:+1: from me

PR #87, opened 2016-06-03
updated app.rb for method_override, port and host

alxngyn:
Fixes issue when using non GET/POST methods on routes.
`enable :method_override` allows the user to use PATCH,DELETE, etc methods and pass it through to Sinatra
not enabling it will only allow GET or PUT methods through

## Changes in this PR. Updated (6/20/16)

- [X] Added `enable :method_override` to environment.rb to utitilze the patch, delete, put, etc methods
- [x] set the `port` to 4567 and `host` to bind to localhost as well
- [X] ~~removed redundant project routes code that should've been deleted from app.rb~~
- [X] ~~added `Iam.run!` to app.rb so the app will serve pages indefinitely when `app.rb` is called~~
- [x] added .bundle/ to .gitignore 

## Testing this PR.

1. `rake` to run the app
2. App should run at localhost:4567
3. `rake spec` should run tests
4. Further testing will come with a PR for the views and routes

### Expected Output.

```
[root@30d5af78747d code]# rake
Migrating to latest
/usr/local/bin/ruby app.rb
== Sinatra (v1.4.7) has taken the stage on 4567 for development with backup from Thin
Thin web server (v1.6.4 codename Gob Bluth)
Maximum connections set to 1024
Listening on 0.0.0.0:4567, CTRL+C to stop
```


app.rb
```diff
@@ -4,6 +4,10 @@
 # IAM - a resource usage metric collection and reporting system
 class Iam < Sinatra::Base
   set :show_exceptions, :after_handler
+  enable :method_override
+  set :port, 4567
```
ElijahCaine: I would put the port and bind variables in a config file and *default* them to these values.

alxngyn: yeah after reading your comments, this is something environments.rb should have and not app.rb

alxngyn:
updated PR with feedback from @ElijahCaine

ElijahCaine:
@alxngyn Running `rake spec` results in the app getting run. Only after quitting do the tests finish. I suggest doing a check before running the app. Something like `If env[TESTING] not set: run tests`

alxngyn:
Hmm this is a bit funky since the tests want the `app.rb` to run the tests but since `app.rb` is configured to run indefinitely, the `rake spec` doesn't know how to stop it. I will look into this more once I get back.

alxngyn:
Thanks to @ElijahCaine for 
```
run! unless ENV['RACK_ENV'] == 'test'
```
This fixed my `rake spec` problem from before

Updated 6/20/16
Cleaned up the commits and rebased with the develop branch.

ElijahCaine:
+1 looks good to me. Glad that line helped, sorry about the slight duplication of efforts.

PR #95, opened 2016-06-08
Elijah/base plugin

### LINKED ISSUE #76: Plugin base class
Base class for ALL plugins so we don't have to rewrite code for each.
Should implement similar methods (register, etc...)

ElijahCaine:
fixes #76

## Changes in this PR.

- [X] Create `BasePlugin` class with common functionality across all plugins.
- [X] Convert `VCPUCount` to use `BasePlugin`.
- [X] Convert `DiskTemplate` to use `BasePlugin`.
- [X] Convert `DiskSize` to use `BasePlugin`.
- [ ] Document the base plugin and how to write a new plugin.
- [X] Migrate `report` and `register` tests over to `BasePlugin` tests (only test what you implement).
- [x] Passes tests.

## Testing this PR.

1. Login to dev environment.
2. Run `rake spec`

### Expected Output.

```
$ rake spec
[... passing tests ...]
Finished in 0.31325 seconds (files took 0.56522 seconds to load)
60 examples, 0 failures
```

@osuosl/devs @Kennric


alxngyn:
Good work on converting all the plugins!
Is there a reason we need a migration file for BasePlugin? 
I think keeping BasePlugin out of our datastores is appropriate.

ElijahCaine:
@alxngyn the reason it has a migration file is so we can test the BasePlugin functionality with the TestingPlugin (the `report` and `register` methods are only tested in `plugins/BasePlugin/spec.rb`.

I actually would really appreciate some feedback on how best to do that. We can try to exclude that from the list of *actual* plug-ins registered somehow. Maybe by putting BasePlugin in a different location in the project (`lib/` for instance).

alxngyn:
hmm I think your approach to putting it in `/lib` would work for now.
We definitely need to get it out of the `/plugins` since scheduler.rb will call it during it's cycle.

ElijahCaine:
@alxngyn moved.

alxngyn:
Perfect. This should help us in the future with other plugins.
+1

PR #97, opened 2016-06-09
Redis to File-Based caching refactor

### LINKED ISSUE #94: Explore using file-based cache instead of Redis
A question was brought up when I was talking with some of the systems folks here at the OSL that I didn't have a great answer to: *"Why don't you just use a file for your cache?"*

As I see it we can store files in memory, load them into the application easily with `JSON.parse()` and write our cache data structure to a file with `JSON.dumps()`.

With that said, let's have a discussion about the pros/cons of using a file instead of Redis. I'll get started:

**Pro:** Less dependencies

**Cons:** Slower (?)

@osuosl/devs @Kennric

ElijahCaine:
fixes #94 

## Changes in this PR.

- [x] Implement `Cache` class with read, write, set, get, and del methods.
- [X] Replace `Redis` usage with new `Cache` object.
- [X] Tests pass.
- [x] Changes documented.
- [x] Remove Redis dependencies.

## Testing this PR.

1. In the dev env run `rake spec`.

### Expected Output.

```
$ rake spec
[ PASSING TESTS! ]

Finished in 0.40368 seconds (files took 0.5816 seconds to load)
70 examples, 0 failures
```

@osuosl/devs @Kennric @alxngyn

### DIFF
--- a/environment.rb
+++ b/environment.rb
@@ -28,7 +28,6 @@ class Iam < Sinatra::Base
   require 'uri'
   require 'openssl'
   require 'json'
-  require 'redis'
 
   # Test stuff
   if env == 'development' || env == 'test'
@@ -55,5 +54,4 @@ class Iam < Sinatra::Base
   end
 
   set :DB, Sequel.connect(settings.database) if defined? settings.database
-  set :REDIS, Redis.new(host: ENV['REDIS_HOST']) if defined? ENV['REDIS_HOST']
 end
--- a/environment.rb
+++ b/environment.rb
@@ -28,6 +28,7 @@ class Iam < Sinatra::Base
   require 'uri'
   require 'openssl'
   require 'json'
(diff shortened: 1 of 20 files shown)


Kennric:
I like it. +1

environment.rb
```diff
@@ -55,5 +54,4 @@ class Iam < Sinatra::Base
   end
 
   set :DB, Sequel.connect(settings.database) if defined? settings.database
-  set :REDIS, Redis.new(host: ENV['REDIS_HOST']) if defined? ENV['REDIS_HOST']
```
alxngyn: Do you think it would be better if we put a `require_relative '../../lib/cache.rb'` in the environment.rb file so it behaves more like the old `@redis`?

ElijahCaine: That's a good idea. I'll try it out.

Issue #83, opened 2016-05-27
Cookbook for production instance


alxngyn:
As we discussed in the sprint meeting, we just need the scheduler be running to collect data while we work on the app. 
- cookbook for the app
- cookbook for postgres(?)
- we probably want backups on the data at some interval

ElijahCaine:
Just to clarify, this cookbook will live in a separate repository on https://github.com/osuosl-cookbooks .

ElijahCaine:
@Kennric Since we are going to use Ansible for deploying, and that will live on a different repository, should this issue still be open?

Kennric:
resolved with ansible deploy and app server cookbooks

PR #100, opened 2016-06-16
Minor changes before the deploy.

ElijahCaine:
## Changes

- Run `ruby app.rb` and `ruby scheduler.rb` with `rake` command.
- When you run `rake` it should automatically start the app and the collector.
- Added `keys` method to cache. Otherwise `scheduler.rb` breaks.

## Testing

```
$ rake spec
[... Tests still pass ...]
```
The testing procedure for PR #68 is also a pretty good way to check this for validity. Run `rake` instead of `ruby scheduler.rb`.
```
[root@c6d98b33fa8a code]# rake
Migrating to latest
/usr/local/bin/ruby app.rb
== Sinatra (v1.4.7) has taken the stage on 4567 for development with backup from Thin
Thin web server (v1.7.0 codename Dunder Mifflin)
Maximum connections set to 1024
Listening on localhost:4567, CTRL+C to stop
```

Refs #83 I guess.

@osuosl/devs @Kennric

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,5 +1,5 @@
 require 'sinatra/base'
-require_relative 'environment.rb'
+require_relative 'environment'
 
 # IAM - a resource usage metric collection and reporting system
 class Iam < Sinatra::Base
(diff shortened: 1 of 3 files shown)


app.rb
```diff
@@ -1,78 +1,18 @@
 require 'sinatra/base'
-require_relative 'environment.rb'
+require 'environment'
 
 # IAM - a resource usage metric collection and reporting system
 class Iam < Sinatra::Base
+  puts 'in Iam class def'
```
Kennric: ?

Kennric:
Not sure why the 'puts' line, but otherwise this looks correct to me. +1

app.rb
ElijahCaine: Sorry, debugging cruft.

Kennric:
+1

Issue #99, opened 2016-06-14
Review spec tests for correctness


ElijahCaine:
There are a few tests, specifically those related to the plugins, which do not actually test what we expect them to. This should be addressed earlier rather than later.

## Procedure for checking test correctness:

1. Run tests.
2. Modify individual test.
3. Run tests again, verify that the test **does not pass**.
  -  If the test still passes, re-write it to **actually test the parameter covered in the test**.
  -  If the test fails, revert that portion of the test back and keep checking.

Tests which are at high-risk for being faulty:
```
.
├── lib
│   └── BasePlugin
│       └── spec.rb
├── plugins
│   ├── DiskSize
│   │   └── spec.rb
│   ├── DiskTemplate
│   │   └── spec.rb
│   ├── RamSize
│   │   └── spec.rb
│   └── VCPUCount
│       ├── migrations
│       └── spec.rb
└── spec
    ├── factories.rb
    ├── models
    │   ├── client_spec.rb
    │   └── project_spec.rb
    ├── routes
    │   ├── app_spec.rb
    │   ├── client_spec.rb
    │   └── project_spec.rb
    └── spec_helper.rb
```

### Note:

Tests should *fail*, not error. If a test gets a syntax error or complains about bad syntax in the test, that is a bad change. The test should still run, just fail.

PR #106, opened 2016-06-29
Thai/review spec tests

athai:
fixes issue #99

## Changes in this PR.

- [X] Review tests for correctness
- [ ] Rewrite tests to actually check for things

## Testing this PR.

Currently N/A.

### Expected Output.

```
$ rake spec
[...]
```

### DIFF
--- a/spec/models/client_spec.rb
+++ b/spec/models/client_spec.rb
@@ -49,9 +49,7 @@ describe 'The Client Model and table' do
     client = Client.create(name: 'Delete Me')
     expect(client).to exist
     expect { client.delete }.to_not raise_error
-    ### Test passes but the item's name and address can still be printed?
-    expect(client).to_not exist
-    print client, ' ', client.name
+    expect(client.exists?).to be false
   end
 
   it 'can update a client' do
--- a/spec/models/client_spec.rb
+++ b/spec/models/client_spec.rb
@@ -49,6 +49,8 @@ describe 'The Client Model and table' do
     client = Client.create(name: 'Delete Me')
     expect(client).to exist
     expect { client.delete }.to_not raise_error
+### Test passes but it doesn't seem like the item is actually deleted?
+### Can still be accessed/printed -- is that supposed to happen??
     expect(client).to_not exist
   end
 
--- a/spec/models/client_spec.rb
+++ b/spec/models/client_spec.rb
@@ -49,9 +49,9 @@ describe 'The Client Model and table' do
     client = Client.create(name: 'Delete Me')
     expect(client).to exist
     expect { client.delete }.to_not raise_error
-### Test passes but it doesn't seem like the item is actually deleted?
-### Can still be accessed/printed -- is that supposed to happen??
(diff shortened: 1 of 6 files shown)


athai:
In the spec tests for DiskSize, DiskTemplate, and VCPUCount, ``(...).to not_exist`` is being used instead of ``(...).to_not exist``. Everywhere else, ``(...).to/to_not exist`` is being used instead. What's the difference?

alxngyn:
So I found out that some of these are passing because they're not being ran at all. I'm fixing them up and will push to this branch when I get through all of them.

Good work @athai !

spec/models/client_spec.rb
```diff
@@ -49,7 +49,9 @@ def app
     client = Client.create(name: 'Delete Me')
     expect(client).to exist
     expect { client.delete }.to_not raise_error
+    ### Test passes but the item's name and address can still be printed?
     expect(client).to_not exist
+    print client, ' ', client.name
```
athai: Is this normal behavior?

alxngyn: I'm 80% sure so. I think the behind the scenes logic is like this:

1. var client is stored in memory to a Client object with name = 'Delete Me'
2. client.delete is ran which is just a SQL execution on the DB
3. the print line is referring data from the variable in the memory and not the DB record.

I think the problem is the `().to exists` and `().to_not exist` which is doing something else than we are expecting.

Kennric: Yes, I think that is correct, Model.delete removes the database entry and returns self (the object you are trying to delete). Probably the way to test this is to try and find the object in the database again and test that it isn't there anymore - or, use Sequel's model method 'exists?' which tests if the object exists in the database (expect client.exists? to eq false)

athai: Is ``expect(client).to_not exist`` logically equivalent to ``expect(client.exists?).to be false``?

alxngyn: From what I could google, they are not equivalent:

> `expect(client).to_not exist` checks if client is falsey (nil or false)
> ```
> From rspec docs:
> expect(actual).to exist        # passes if actual.exist? and/or actual.exists? are truthy
> ```
> https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers

> `expect(client.exists?).to be false` checks if the result from `client.exists?` is false (not falsey)
> ```
> From rspec docs:
> expect(actual).to be false     # passes if actual == false
> ```
> https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers

> 
> `client.exists?` should return true or false depending if client exists
> ```
> From rspec docs: 
> The exist matcher is used to specify that something exists
> (as indicated by #exist? or #exists?):
> ```
> https://www.relishapp.com/rspec/rspec-expectations/v/2-5/docs/built-in-matchers/exist-matcher

Kennric:
That looks pretty good, tests are running, and fail when I modify them slightly, as they should. +1

alxngyn:
The tests still pass but the fixes work as they should. Good work. +1

PR #108, opened 2016-06-30
meeting notes for Thursday, July 30, 2016.

ElijahCaine:
@alxngyn @Kennric @athai meeting notes from earlier today. let me know if I missed anything we should record.


alxngyn:
++

PR #112, opened 2016-07-03
Implemenation docs

ElijahCaine:
## What is this?

In an effort to help some of the new OSL Academy students (and future community developers that may be interested in helping) I thought it might be nice to outline what many of the important files in IaM do.

Each section of this document outlines what that file / directory does. It's more or less a brain-dump to hopefully answer some basic questions while I'm not around.

I encourage anybody who is able to fill this document out further to do so. It doesn't need to be perfect it just needs to be able to answer simple questions like "What does this file do?" and "Where is X exactly?" when you're not around.

## Changes in this PR.

- [x] Added skeleton docs for the IaM Implementation.
- [X] Added docs related to the work I have done.
- [ ] Completed the implementation docs.

## Testing this PR.

1. Go to the `docs` directory.
2. In a python virtualenv install `sphinx` and run `make html`.
3. In the docs' build/html directory run `python -m SimpleHTTPServer` and go to `localhost:8000` in your browser.
4. Navigate to the Implementation docs.

@alxngyn @Kennric @athai Please add to these docs and merge them when you think they are good enough.


alxngyn:
Can we just document at the top of the file themselves? 
I feel like if we merge this in, we will have to manage this whenever we make a change in the code which is more work.

docs/source/implementation/index.rst
```diff
@@ -0,0 +1,390 @@
...
+
+  * **app**: Declares the base docker image including *which file to use to
+    build the environment*, some *environment variables* to set, and which
+    directory to set as the *current working directory*.
+  * **dev**: Is *the actual container we use*. It declares *which command to run
+    by default*, that it would like to be linked to **postgres**, which *ports to
+    expose* when it's being run, and which *volumes to mount* to the running
+    container.
+  * **postgres**: is a container which allows us to test our code on an actual
+    postures database.
```
athai: postures?

ElijahCaine:
@osuosl/devs I would like to know if we want to continue and maintain this document or put it to the side. If we really think it's overkill that's fine. I still firmly believe that it will be useful but I'm also democratically minded so let's vote.

alxngyn:
@ElijahCaine While I think the docs are great. I think to **read** them you shouldn't need to run commands like `make html` and `python -m simpleHTTPServer`.  It should be just available. If we do go this method there should be a branch for github pages where it is hosted.

ElijahCaine:
@alxngyn That's not a bad idea. Using GH pages to host repo docs is good. Alternatives to that, which also make this doc more accessible, include:

- Using a GH repo Wiki.
- Telling people to read the file on GH which will format it correctly out of the box.

I'm not against GH pages, but that also involves someone running a script (or manual process) to update the GH pages branch which can be tricky / it may get updated.

PR #114, opened 2016-07-06
Cthalmann/add tests

cthalmann:
Added util tests and methods for iam.

## Changes in this PR.
  Adds methods and tests for finding the minimum, maximum, and average values in an array of hashes. Also adds a test to make sure all values in an array of hashes have positive integer values.

@osuosl/devs


Kennric:
The tests look pretty good so far, but I think they should be broken up by method. I think a describe block should hold all the tests for a particular method:
`
describe 'the util.average method' do
  it 'returns what we expect' do
    expect(the right thing).to eq(the right answer)
  end
  it 'does another thing' do
    expect(another thing).to eq(the right answer)
  end
end

describe 'the util.max method' do
  it 'has some tests' do
    etc
  end
end
`

spec/util/util_spec.rb
```diff
@@ -1,28 +1,95 @@
...
+        node: 'amahi.osuosl.org', value: -2, active: true },
+      { id: 3, node_resource: nil,
+        created: Time.parse('2016-07-01 21:43:28 +0000'),
+        node: 'answers.ros.osuosl.org', value: 80, active: true },
+      { id: 4, node_resource: nil,
+        created: Time.parse('2016-07-01 21:43:27 +0000'),
+        node: 'cthalmann.osuosl.org', value: 4, active: true }
+    ]
+
+    expect(Util.is_positive_integer_value(@data_set_false)).to eq(false)
```
Kennric: I'm not sure how useful this proposed method is.... @alxngyn @ElijahCaine  is it worth having a separate method to do this data validation for us? Alternatives: just rely on validations before the data is even stored (in collectors), or have these util methods simply raise and error on bad values

util.rb
```diff
@@ -4,58 +4,48 @@
...
-          flag = false
-        end
+        flag = false if data[i][:value] < 0
       else
         flag = false
       end
       i += 1
     end
-    return flag
+    flag
```
mathuin: I'm curious -- what inspired you to use a flag variable instead of returning false at lines 43 and 45?

cthalmann: I was taught to only use multiple return statements in code if I'm checking something right at the beginning, such as line 7 when checking to see if the list is empty. Otherwise, I was taught to not use multiple return statements, reasoning being that it is easier to debug.

Kennric:
I think the tests are looking pretty good.

@alxngyn - what do you think of the code, and the utility of the utility functions?

mathuin:
Today I learned about `Enumerable`.  http://ruby-doc.org/core-2.3.1/Enumerable.html

Using this mixin would give you `max` and `min` as well as `count` which is half of what you need to calculate the average.  The sum could come from `data.reduce(:+)` which is in the `reduce` example.

util.rb
mathuin: While it's true that procedural languages work well with a single entry and a single exit, that's not really the case with Ruby.  Ruby's idiomatic style is pretty different from C, Perl, Python, Go and darn near every other language I can think of right now. :-)  https://github.com/bbatsov/ruby-style-guide is a reasonable description of idiomatic Ruby style in the wild, and it also does a great job of showing the potential dangers in using return as you might do in other languages -- something I didn't know until just now!

mathuin:
Much shorter and sweeter, and passes -- 85 examples, 0 failures.  :thumbsup:

alxngyn:
Tests are solid and work for me. The only thing I would criticize is the use spaceship operators (<=>).

Kennric:
@cthalmann looks pretty good - I have one more change for you, and it requires a bit of git rebasing.

We've merged another pr that adds a util file in lib/util.rb - I think we should add this util class to that file, and update any require statements as necessary. To avoid confusion, perhaps we should rename this class DataUtil or something to make it clear that these methods are for massaging data.

You'll need to rebase your branch from develop, then move things around.

mathuin:
I tried to run tests on your branch and got the following error:

$ docker-compose run --service-ports --rm dev bash
bash: /data/code/dockerfiles/startup.sh: No such file or directory
[root@a8f838867674 code]# exit
$ 

Any idea why that happened?

Kennric:
I think that's got it. I'd like @hsolorza, @elenatorrey, @lo9key to also review, just to be familiar with what you did. When you have +1 from each of them, go ahead and merge.

lo9key:
* Is DataUtil the best name for the class?  It seems very specific but also pretty vague.
* I would change the variable name on the sample data to something other than @data_max_80, because it's being used in all of the tests instead of just to find the maximum.

These are just my opinions, overall it looks good. It passes the tests, I see no flaws in your logic, and rubocop is happy. +1

hsolorza:
Looks good.
+1

elenatorrey:
+1

PR #115, opened 2016-07-06
Levi/util spec tests

lo9key:
Adds utility tests and methods

## Changes in this PR.

- [X] Find min, max, avg value
- [X] Find number of days included in a data set
- [X] Uses sample data.
- [ ] Uses real data.

## Testing this PR.

1. Create a docker container `docker-compose run dev bash`
2. Run tests `rake spec`
3. Run linter `rubocop <filename>`
4. More development needed before methods can actually be used.  (running on sample data now.)

### Expected Output.

```
$rake spec
The DataUtil class tests
  the max_value method
    returns the max value
    returns 0 for max value if array is empty
  the min_value method
    returns the min value
    returns 0 for min value if array is empty
  the average_value method
    returns the average value
    returns 0 for average of values if array is empty
  the days_in_range method
    properly handles a range of dates with identical times of day
    returns correct number of days in a range
    returns 1.0 for a single day
    properly handles leap years
    returns nil if called with invalid data

Finished in 0.50804 seconds (files took 2.04 seconds to load)
93 examples, 0 failures

[root@5d5df056f44c code]# rubocop lib/util.rb 
lib/util.rb:93:3: C: Assignment Branch Condition size for days_in_range is too high. [20.86/15]
  def self.days_in_range(data)
  ^^^
lib/util.rb:93:3: C: Method has too many lines. [13/10]
  def self.days_in_range(data) ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
1 file inspected, 14 offenses detected

[root@d886978cac01 code]# rubocop util.rb 
Inspecting 1 file
Offenses:
util.rb:40:3: C: Assignment Branch Condition size for days_in_range is too high. [20.86/15]
  def self.days_in_range(data)
  ^^^
util.rb:40:3: C: Method has too many lines. [13/10]
  def self.days_in_range(data) ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
1 file inspected, 2 offenses detected

[root@5d5df056f44c code]# rubocop spec/util/util_spec.rb 
Inspecting 1 file
.
1 file inspected, no offenses detected
```

I believe the above caution flags on my days_in_range method are justified due to how complicated timestamp testing can be.  The other 12 'C' offenses are from existing code in the util file.

@osuosl/devs


mathuin:

util.rb
```diff
@@ -0,0 +1,61 @@
...
+      return nil if l.class != Time
+      # dates cannot be in the future
+      return nil if l > now
+      earliest = l if l < earliest
+      latest = l if l > latest
+    end
+    # latest-earliest is number of seconds between two dates
+    # there are 60*60*24 seconds every day (leap years? DST?)
+    # round up to whole days?  Maybe not?
+    ((latest - earliest) / (60 * 60 * 24)).ceil
```
mathuin: It's true that this answer won't cover leap years nor will it account for daylight savings times.  I'm not sure whether rounding matters.  However, if I had to calculate the number of days between two Time values, I'd probably look at http://stackoverflow.com/questions/9608116/number-of-days-between-two-time-instances -- and while the accepted answer is the rounded version of yours, I personally like the simpler version of the second answer.

The `Enumerable` mixin may also be helpful here to identify the earliest and latest times.

lo9key: Thanks for the feedback @mathuin.

I've been thinking about it and I suspect this should work fine with DST and leap years, because the time stamps are in UTC and (should) play nice  with the additional day of a leap year.

I'm not sure how to test DST, but I can definitely add a leap year test by choosing days on either side of a known leap year.  I'll add that test case when I get in tomorrow.

I think you're probably right, a float can be rounded later in the program, but returning an int in the method reduces precision.  I'll check with @ramereth and @Kennric to confirm.

mathuin:
Awesome!  87 examples, 0 failures.  :thumbsup:

alxngyn:
Cant we move `Util.rb` to `lib/`? `lib/` is a place for reference files that don't directly interact with the app. You'll need to fix the reference location in `util_spec.rb`.  Aside from that, every looks good.

lo9key:
@alxngyn @Kennric @ramereth I believe this PR is finished and ready for review.

alxngyn:
+1 I like the date testing implementation from what I've seen

alxngyn:
@lo9key Can you handle the merge conflicts and retest? I'd like to get your tests in.

lo9key:
@alxngyn @cthalmann @Kennric I merged my function with thal's and made sure it passed `rake spec` and `rubocop`.  See my PR notes above about a caution on the length of my method.

I believe this PR is ready for review.

cthalmann:
:+1:

alxngyn:
+1

PR #116, opened 2016-07-06
tests for iam

elenatorrey:
fixes issue #xx

## Changes in this PR.

- [X] fixes the avg, min, max methods

## Testing this PR.

1. Do X.
2. Do Y.
3. See Z.

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/util.rb
+++ b/util.rb
@@ -23,32 +23,19 @@ class Util
   end
 
   def self.avg_val(data)
-    # return average value
-    return 0 if data.empty?
-    sum = 0
-    data.each do |item|
-      sum += item[:value]
-    end
-    sum / data.length
+    # returns the average of values
+    data.map { |a| a[:value] }.reduce(:+) / data.length
   end
 
   def self.min_val(data)
     # return the minimum of values
     return 0 if data.empty?
-    min = data[0][:value]
-    data.each do |item|
-      min = item[:value] if item[:value] < min
-      return min
-    end
+    data.min { |a, b| a[:value] <=> b[:value] }[:value]
   end
 
   def self.max_val(data)
     # return the maximum of values
     return 0 if data.empty?
-    max = data[0][:value]
-    data.each do |item|
-      max = item[:value] if item[:value] > max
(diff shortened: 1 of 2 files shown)


elenatorrey:
@Kennric

mathuin:

util.rb
```diff
@@ -0,0 +1,54 @@
...
+    sum / data.length
+  end
+
+  def self.min_val(data)
+    # return the minimum of values
+    return 0 if data.empty?
+    min = data[0][:value]
+    data.each do |item|
+      min = item[:value] if item[:value] < min
+      return min
```
mathuin: Are you sure this line is in the right place?  Compare it to the following method.

util.rb
```diff
@@ -0,0 +1,54 @@
...
+class Util
+  def self.val_id(data)
+    # checks if the hashes are empty
+    # checks if the values of id are fixed numbers
+    return 0 if data.empty?
+    data.each do |item|
+      id = item[:id]
+      return false if id.class != Fixnum
+    end
+    true
```
mathuin: Is this line necessary?  What happens if you remove the `true` in this method and the method that follows?  Do the methods return the same results?  Which one is more idiomatic?

util.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def self.id_not_neg(data)
+    # checks if the id is not negative
+    data.each do |item|
+      return false if item[:id] <= 0
```
alxngyn: this function will return false if the item is 0

elenatorrey:
@Kennric @alxngyn

alxngyn:
@elenatorrey there are some still pressing concerns from Mathuin and I

elenatorrey:
Since docker breaks everything I can't fix it right now:(

alxngyn:
@elenatorrey 
Can we fix the merge conflicts and retest? I think we can use your validation tests in the future.

PR #117, opened 2016-07-06
Added tests and methods to the Util class

hsolorza:
fixes issue: 
Creates tests and methods for the Util file.

## Changes in this PR.

- [X] Minimum Value
- [X] Maximum Value
- [X] Average Value 

## Testing this PR.

- Run with  _sudo docker-compose run dev bash_

- Then _rake spec_

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

Included test for finding the maximum value.
Included test for finding the minimum value.
Included test for calculating the average of the values.
Included tests to verify valid Active and Node Resource entries.

### DIFF
--- a/util.rb
+++ b/util.rb
@@ -1,0 +1,23 @@
+require 'sinatra/base'
+require_relative 'environment'
+
+# IAM - a resource usage metric collection and reporting system
+class Util
+  def self.max_value(data)
+    # find the max Value
+    return 0 if data.empty?
+    data.max { |item, item2| item[:value].abs <=> item2[:value].abs }[:value]
+  end
+
+  def self.min_value(data)
+    # find the lowest Value
+    return 0 if data.empty?
+    data.min { |item, item2| item[:value].abs <=> item2[:value].abs }[:value]
+  end
+
+  def self.avg_mean_value(data)
+    # find the average(Mean) of the Values
+    return 0 if data.empty?
+    data.map { |item| item[:value].abs }.reduce(:+) / data.count
+  end
+end
--- a/spec/util/util_spec.rb
+++ b/spec/util/util_spec.rb
@@ -1,4 +1,6 @@
 require File.expand_path '../../spec_helper.rb', __FILE__
+require_relative '../../util.rb'
+require 'time'
 
 describe 'The Utility tests' do
   def app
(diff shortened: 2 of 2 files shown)


mathuin:

util.rb
```diff
@@ -0,0 +1,60 @@
...
+  def self.active_bool(data)
+    # insure that Active is either True or False
+    data.each do |item|
+      if item[:active] == true || item[:active] == false
+        # do nothing
+      else
+        return false
+      end
+    end
+    true
```
mathuin: In this method you explicitly return true but in the following method you don't.  Is there any particular reason?  Do both work?  If they both work, is one approach better than the other?

util.rb
```diff
@@ -38,20 +38,15 @@ def self.avg_mean_value(data)
...
-      end
+      return false if item[:active] != true && item[:active] != false
     end
-    true
   end
 
   def self.node_resrc(data)
     # insure node resource is an int or nil
     data.each do |item|
       if (item[:node_resource].is_a? Integer) || item[:node_resource].nil?
```
mathuin: The trick you use in the method above this one could also be used here!

util.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def self.avg_mean_value(data)
+    # find the average(Mean) of the Values
+    data.map { |item| item[:value].abs }.reduce(:+) / data.count
+  end
+
+  def self.active_bool(data)
+    # insure that Active is either True or False
+    data.each do |item|
+      return false if item[:active] != true && item[:active] != false
```
alxngyn: i believe this line will always return false unless things in ruby can be neither true or false.

util.rb
```diff
@@ -0,0 +1,40 @@
...
+class Util
+  def self.max_value(data)
+    # find the max Value
+    highest = data.max { |item, item2| item[:value].abs <=> item2[:value].abs }
+    highest[:value]
+  end
+
+  def self.min_value(data)
+    # find the lowest Value
+    lowest = data.min { |item, item2| item[:value].abs <=> item2[:value].abs }
```
alxngyn: @Kennric, should we use the spaceship operator? 
I know Codecademy teaches it but it is rarely used in practice/industry from my experience. It also makes the code confusing.

alxngyn:
@hsolorza Is it safe to assume that this PR is no longer needed and can be closed?

Issue #81, opened 2016-05-20
REFACTOR: DiskSize for Disk Number


alxngyn:
Refactor Redis/Collector and DiskSize plugin to implement each disk value available

PR #119, opened 2016-07-11
Refactored disk_size plugin model to take array of values

alxngyn:
fixes issue #81, should allow us to have finer detail on disk_sizes in the future. The only caveat is that max disk count is currently at 3. We can change in the future if we find a need more than that.

## Changes in this PR.

- [X] Refactored the disk_size plugin's data model
- [X] Updated disk_size's tests for the new data model

## Testing this PR.

1. Start IAM in docker
2. `rake spec`

### Expected Output.

```
$ rake spec
...

DiskSize plugin
  .store method
    does not fail with valid data
No disk_sizes information for badnode: {"disk_size"=>"[10, 20]", "active"=>true}
    properly stores all disk sizes when storing in DB
...
Finished in 0.26475 seconds (files took 0.74526 seconds to load)
88 examples, 0 failures
```

@Kennric @subnomo @athai @ElijahCaine

### DIFF
--- a/plugins/DiskSize/migrations/001_migrations.rb
+++ b/plugins/DiskSize/migrations/001_migrations.rb
@@ -8,7 +8,10 @@ Sequel.migration do
       foreign_key   :node_resource
       Time          :created, null: false
       String        :node
-      Integer       :value
+      Integer       :disk_count
+      Integer       :disk1_size, default: 0
+      Integer       :disk2_size, default: 0
+      Integer       :disk3_size, default: 0
       Boolean       :active
     end
   end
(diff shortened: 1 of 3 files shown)


plugins/DiskSize/migrations/001_migrations.rb
```diff
@@ -8,7 +8,10 @@
       foreign_key   :node_resource
       Time          :created, null: false
       String        :node
-      Integer       :value
+      Integer       :disk_count
```
Kennric: This rename makes sense from a plugin perspective, but if all measurement tables use the same field name for their measured value, then it will be easier to write utility methods to work with report data in a generic way. So I vote for using 'value', or another generic name for the measurement field

alxngyn:
After some discussion of needing an separate values vs summed values, we decided it would probably be simplier for now if we kept the summed values. We can revisit this PR if we ever find a need for the separate values in the future.

Issue #120, opened 2016-07-11
fix bundle install issues


alxngyn:
Running just `rake` is starting become a dependency nightmare. Running `bundle install` just installs duplicate gems and ruby will complain about them. 
Let's pin a specific version for each of the gems so we don't have to waste time every time.

ElijahCaine:
Adding Sprint milestone so we can wrap this issue up.

PR #130, opened 2016-07-18
Pinned all gems to a version on a working build

alxngyn:
fixes issue #120 

## Changes in this PR.

- [X] Pinned all gems to a specific version from a working build

## Testing this PR.

1. Do docker
2. If this is your first run in this branch, do a bundle install and it should install all the specific gems
3. Do `rake`
4. It should just work and not complain about gem versions/not found/etc

### Expected Output.

```
$ rake spec
[ ...successful output goes here ]


ElijahCaine:
This PR is out of date, or will very quickly become out of date with #145 getting merged.
Do we think this should get merged or should we scrap it?

alxngyn:
I think we should pin them and upgrade each gems version if we need a specific feature or security update. 
@Kennric thoughts?

Kennric:
@alxngyn @ElijahCaine yeah, if we are running into version issues, we should pin and re-pin as needed

ElijahCaine:
@alxngyn Can we please merge this soon? If you don't want to do it I will, I don't want to step on your toes though.

hsolorza:
+1 lgtm

ElijahCaine:
Tests pass when you run `docker-compose build --no-cache --pull && docker-compose up dev` so I'm gonna give this a +1.

Issue #137, opened 2016-08-01
Database resource plugin - store [5] 


alxngyn:
store method for the db plugin

Issue #136, opened 2016-08-01
Database resource plugin - migration [1]


alxngyn:
migrations file for the DB plugin

PR #143, opened 2016-08-09
DBSize Plugin - WIP

alxngyn:
fixes issue #137 #136 
**Plugin will not work since the db collector is currently in dev.**

## Changes in this PR.

- [X] Created DBSize plugins.rb
- [X] Created DBSize migrations
- [X] Created spec.rb 

## Testing this PR.

1. Do docker-compose.
2. Do rake spec.
3. See plugin tests pass.

### Expected Output.

```
$ rake spec
[...]
```


plugins/DBSize/plugin.rb
```diff
@@ -0,0 +1,39 @@
...
+    node_info = @cache.get(fqdn)
+    db_key = 'Data Base Size in Bytes'
+    # Check for valid data
+    if node_info[db_key].nil? || node_info[db_key] == ''
+      raise "No DBSize information for #{fqdn}\n"
+    end
+
+    # Insert data into disk_size_measurements table
+    @@database[@@table].insert(
+      node:          fqdn,
```
ElijahCaine: I don't know if databases have FQDNs. The way it's currently setup each database is stored in the cache as such:
```
"database_name":"######"
```
where `#####` is the database size in bytes. A full FQDN would be something like `mysql://username:password@db_host/database_name`.
If you think that's a better way to store the database in the cache we can make that change, just a heads up.

plugins/DBSize/migrations/001_migrations.rb
```diff
@@ -0,0 +1,15 @@
+# DiskSize's migration file
+
+Sequel.migration do
+  transaction
+  change do
+    create_table(:db_size_measurements) do # Plugin's DB model
+      primary_key   :id
+      foreign_key   :node_resource
+      Time          :created, null: false
+      String        :node
```
ElijahCaine: We should really not call these nodes. Nodes is more of a Cloud VM term. These should probably be called `databases` or `dbs`.

plugins/DBSize/plugin.rb
```diff
@@ -0,0 +1,39 @@
+require 'sequel'
+require_relative '../../lib/BasePlugin/plugin.rb'
+require_relative '../../environment.rb'
+require_relative '../../models.rb'
+
+# Disk Sizes data plugin
+class DBSize < BasePlugin
+  def initialize
+    @@name = 'DBSize'
+    @@resource_name = 'node'
```
ElijahCaine: `'node'` -> `'dbs'`

plugins/DBSize/plugin.rb
```diff
@@ -0,0 +1,39 @@
...
+    db_key = 'Data Base Size in Bytes'
+    # Check for valid data
+    if node_info[db_key].nil? || node_info[db_key] == ''
+      raise "No DBSize information for #{fqdn}\n"
+    end
+
+    # Insert data into disk_size_measurements table
+    @@database[@@table].insert(
+      node:          fqdn,
+      value:         node_info[db_key],
```
ElijahCaine: `node_info` -> `db_info`

plugins/DBSize/plugin.rb
ElijahCaine: `node:` -> `db:`

plugins/DBSize/spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  # Store method
+  describe '.store method' do
+    before(:all) do
+      DBSize.new.register
+      @cache = Cache.new(ENV['CACHE_DIR'])
+    end
+
+    before(:each) do
+      @cache.set('test_db', db_key: "123456.0", active: true)
```
ElijahCaine: If you want the cache to look like this we can try to do that, but the current cache is set as
```
@cache.set(var[:"DB Name"], var[:"Data Base Size in Bytes"])
```
where `var[:"DB Name"]` is the database name and `var[:"Data Base Size in Bytes"]` is the size in bytes. It's pretty bare-bones but it's all the info we have to work with for now.

plugins/DBSize/plugin.rb
alxngyn: so should we change `fqdn` to `db_host`?

plugins/DBSize/spec.rb
alxngyn: `db_key` is the same i set it in the `before(:all)` clause because it was too long line-wise
 and the 'test_db' is just a temp name for the test

plugins/DBSize/plugin.rb
```diff
@@ -0,0 +1,39 @@
+require 'sequel'
+require_relative '../../lib/BasePlugin/plugin.rb'
+require_relative '../../environment.rb'
+require_relative '../../models.rb'
+
+# Disk Sizes data plugin
+class DBSize < BasePlugin
+  def initialize
+    @@name = 'DBSize'
```
ElijahCaine: In the other plugins iirc these variables also need to be set *before* the initialize block. I don't remember why I had to do that but just make sure this won't bite us later.

plugins/DBSize/plugin.rb
```diff
@@ -0,0 +1,39 @@
...
+      raise "No DBSize information for #{db_host}\n"
+    end
+
+    # Insert data into disk_size_measurements table
+    @@database[@@table].insert(
+      db:            db_host,
+      value:         db_info[db_key],
+      active:        1,
+      created:       DateTime.now,
+      node_resource: @@database[:node_resources].where(name: db_host).get(:id))
```
ElijahCaine: This isn't a node. Don't call it a `node_resource`.

plugins/DBSize/spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      puts @db_table.where(db: 'test_db')
+      # Check that store actually stored the node
+      expect(@db_table.where(db: 'test_db')).to_not be_empty
+    end
+
+    it 'fails when not passed node name' do
+      # This is bad plugin usage that should actually crash
+      expect { DBSize.new.store }.to raise_error(ArgumentError)
+    end
+
```
ElijahCaine: Should we have a test that `does fail with invalid data`?

plugins/DBSize/spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    after(:each) do
+      @cache.del('test_db')
+      @cache.write
+      @db_table.where(db: %w(test_db test_db2)).delete
+    end
+
+    it 'does not fail with valid data' do
+      # Store cached nodes in DB, no error
+      expect { DBSize.new.store('test_db') }.to_not raise_error
+      puts @db_table.where(db: 'test_db')
```
ElijahCaine: Is this supposed to be there?

plugins/DBSize/plugin.rb
alxngyn: All the other plugins use this order. I can't find what you're talking about.

plugins/DBSize/plugin.rb
alxngyn: I thought a node resource isn't the same as a vm node. My understanding is that a project can have many node resources like vms, ftps, DBs. It's probably a bad labeling we had from the start.

plugins/DBSize/spec.rb
alxngyn: Working on it

plugins/DBSize/plugin.rb
```diff
@@ -0,0 +1,43 @@
...
+  def store(db_host)
+    # Pull node information from cache as a ruby hash
+    db_info = @cache.get(db_host)
+    db_key = 'Data Base Size in Bytes'
+
+    # Check for valid data
+    if db_info[db_key].nil? ||
+      db_info[db_key] == '' ||
+      db_info[db_key].is_a?(Numeric) == false
+        puts "error raised"
```
hsolorza: Is this _puts_ supposed to be here?

plugins/DBSize/plugin.rb
```diff
@@ -0,0 +1,43 @@
+require 'sequel'
+require_relative '../../lib/BasePlugin/plugin.rb'
+require_relative '../../environment.rb'
+require_relative '../../models.rb'
+
+# Disk Sizes data plugin
```
hsolorza: _DiskSize_ is one word. Also needs an apostrophe.

hsolorza:
tests pass, looks good    +1

plugins/DBSize/plugin.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    # Insert data into db_size_measurements table
+    @@database[@@table].insert(
+      db:            db_host,
+      value:         db_info[db_key],
+      active:        1,
+      created:       DateTime.now,
+      node_resource: @@database[:node_resources].where(name: db_host).get(:id))
+  rescue => e                        # Don't crash on errors
+    STDERR.puts "#{e}: #{db_info}" # Log the error
```
hsolorza: Is there going to be a line to log this error?

ElijahCaine:
This looks good to me. :+1:

plugins/DBSize/migrations/001_migrations.rb
```diff
@@ -0,0 +1,15 @@
+# DiskSize's migration file
```
hsolorza: Typo: _DBSize's_

hsolorza:
1+ lgtm

Issue #77, opened 2016-05-20
Logging implementation


alxngyn:
Nice to have feature,
* research which logging gem would be best
* create a config where users can dictate where the log file goes

matthewrsj:
A couple options I've found:

* https://github.com/TwP/logging
* https://github.com/colbygk/log4r

They are pretty similar, but the first option seems to require slightly less work:

From `logging` documentation:
```ruby
require 'logging'

logger = Logging.logger(STDOUT)
logger.level = :warn

logger.debug "this debug message will not be output by the logger"
logger.warn "this is your last warning"
```

From `log4r` documentation:
```ruby
require 'log4r'
include Log4r

# create a logger named 'mylog' that logs to stdout
mylog = Logger.new 'mylog'
mylog.outputters = Outputter.stdout

# Now we can log.
def do_log(log)
  log.debug "This is a message with level DEBUG"
  log.info "This is a message with level INFO"
  log.warn "This is a message with level WARN"
  log.error "This is a message with level ERROR"
  log.fatal "This is a message with level FATAL"
end
do_log(mylog)
```

alxngyn:
I vote for logging and how it is easy to setup and use. The setup/use is also similar to everything else we do.

ElijahCaine:
As long as we can specify in a configure file where we are logging to I think both are adequate. If they're about even I'd go with the one that requires fewer dependencies or the one that is more heavily maintained.

hsolorza:
Logging added #146 and #145

PR #145, opened 2016-08-12
Create logging 

hsolorza:
fixes issue #77 

## Changes in this PR.

- [X] Creates a log
- [x] Tests the log's output

## Testing this PR.

1. In a container, run _rake spec_

### Expected Output.

```
$ rake spec
Finished in 0.66719 seconds (files took 1.05 seconds to load)
94 examples, 0 failures
```

@osuosl/devs


ElijahCaine:
This looks good so far. Something I would consider adding is a `log` component to the `app` so that when we can do something like `app.log(message)` or add a few different levels of logging so we can do `app.warn(message)` and `app.debug(message)` and `app.panic(message)` for example.

Basically look around the application and see if we can add a logging component to some of the objects/classes we're using.

Let me know if that makes sense. So far this looks good :smile:

Rakefile
```diff
@@ -37,7 +37,7 @@ end
 # rake spec
 RSpec::Core::RakeTask.new(:spec) do |t|
   ENV['RACK_ENV'] = 'test'
-  t.pattern = Dir.glob('{lib,spec,plugins}/**/*spec.rb')
+  t.pattern = Dir.glob('{lib,spec,plugins,logging}/**/*spec.rb')
```
Kennric: Does this actually find and run your log spec?

hsolorza: I think so. Without this line, the test wont work.

hsolorza:
@ElijahCaine like this  https://github.com/osuosl/iam/pull/146 ?

ElijahCaine:
@hsolorza Yeah, that's actually not too bad.

logging/logs.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require 'logging'
+require 'environment.rb'
+
+# Creates logfile with the level 'debug'
+#   that appends to log_file.log
+log = Logging.logger['logfile']
+log.level = :debug
+log.add_appenders \
+   Logging.appenders.file('log_file.log')
```
alxngyn: can we add the logs to a logs/ directory? I feel like our home directory is already busy with collectors, schedulers, app, etc.

hsolorza: Sure, this was just a dummy file as I wasn't sure where you wanted the actual log file to be located.

logging/logs.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require 'logging'
+require 'environment.rb'
+
+# Creates logfile with the level 'debug'
+#   that appends to log_file.log
+log = Logging.logger['logfile']
+log.level = :debug
+log.add_appenders \
+   Logging.appenders.file('logging/log_file.log')
```
ElijahCaine: This should be an environment variable not hard-coded as `logging/log_file.log`.
It can fall-back on that but this shouldn't be the only behavior.

ElijahCaine:
Tests pass and the code looks good.
Once you replace that hard-coded path with an environment variable (and fallback on that hard-coded path) then this will get a +1 from me.

alxngyn:
+1 this feature will help keep our logs organized once we start running

hsolorza:
@ElijahCaine  Did you want me to move this to the environment.rb?

logging/logs.rb
```diff
@@ -0,0 +1,19 @@
+# Types of log statements:
+#   log.debug "a very nice little debug message"
+#   log.info "things are operating nominally"
+#   log.warn "this is your last warning"
+#   log.error StandardError.new("something went horribly wrong")
+#   log.fatal "I Die!"
+
+require 'logging'
+require 'environment.rb'
+@file = ENV['logging/log_file.log']
```
ElijahCaine: So this would require us to have an environment variables called `logging/log_file.log` which would be set to the path of our log file. The more standard way to do this is to name the variable `LOG_FILE_PATH` which we would set in `dockerfiles/app.env.dist` to `LOG_FILE_PATH=/path/to/log_file.log`.

logging/logs.rb
```diff
@@ -0,0 +1,19 @@
...
+# Creates logfile with the level 'debug'
+#   that appends to log_file.log
+log = Logging.logger['logfile']
+log.level = :debug
+log.add_appenders \
```
ElijahCaine: This should read more like `log.add_appenders Logging.appenders.file(ENV['LOG_FILE_PATH'].is_set? ? ENV['LOG_FILE_PATH'] : 'logging/log_file.log')` which translates to `use ENV['LOG_FILE_PATH'] if it's set, otherwise default to 'logging/log_file.log'`

ElijahCaine:
@hsolorza As it stands it's fine to keep it where it is. I was just curious why the code was organized the way it is. I do have more feedback, but this is looking good.

logging/logs.rb
Kennric: I second this.

Kennric:
Looking good, I think if you address @ElijahCaine's comment above, I am ready to approve

Kennric:
+1

ElijahCaine:
@hsolorza This looks great. :+1: from me :smile:

Issue #102, opened 2016-06-27
Client, project and node views


alxngyn:
This will be a 3 part PRs. One for each of the client, project and node views.
I'll try to be thorough with each piece but we'll figure out the details along the way.

alxngyn:
@subnomo I'm assigning you to the project views. It should be similar to PR #103 
Lets sit down for 30mins/hour to talk about implementing tomorrow or next week.

alxngyn:
although we still have work to do on all the views, I would call this PR is now done

PR #150, opened 2016-08-23
Project views

hsolorza:
fixes issue #102 

## Changes in this PR.

- [X] Adds create, edit, and view options for projects

## Testing this PR.

1. _sudo docker-compose run --service-ports dev bash_
2. _rake migrate_
3. create temp file at same level as the rakfile with: 

```
require './environment.rb'
require 'models'

# generate data
 (0...25).each do
   o = [('a'..'z'), ('A'..'Z')].map(&:to_a).flatten
   name = (0...10).map { o[rand(o.length)] }.join

   new_client = Client.create(name:  name,
                              contact_name:  'bob',
                              contact_email:  'bob@example.com',
                              description:  'blah')

   name = (0...10).map { o[rand(o.length)] }.join

   new_project = Project.create(name: name,
                                client_id: 1 + rand(new_client.id),
                                resources: 'node,thingy',
                                description: 'blah')

   name = (0...10).map { o[rand(o.length)] }.join

   NodeResource.create(project_id: 1 + rand(new_project.id),
                       name: name,
                       type: 'ganeti',
                       cluster: 'cluster1.osuosl.org',
                       created: DateTime.now,
                       modified: DateTime.now)
 end
```

4.run the file
5. _rake_

**Index (top part only):**

screenshot from 2016-08-23 13 45 52

**Show:**

screenshot from 2016-08-23 13 46 14

**Edit:**

screenshot from 2016-08-23 13 46 30

**Create:**

screenshot from 2016-08-23 13 46 46

@osuosl/devs

### DIFF
--- a/views/projects/create.erb
+++ b/views/projects/create.erb
@@ -1,5 +1,5 @@
 <div class="container">
-  <form action='/projects' name="createProject" method='post' onsubmit="returnt validateForm()">
+  <form action='/projects' name="createProject" method='post' onsubmit="return validateForm()">
     <table class='table'>
     <thead>
       <tr>
--- a/views/projects/create.erb
+++ b/views/projects/create.erb
@@ -1,0 +1,49 @@
+<div class="container">
+  <form action='/projects' name="createProject" method='post' onsubmit="returnt validateForm()">
+    <table class='table'>
+    <thead>
+      <tr>
+        <th>Name</th>
+        <th>New Name</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td>Project name</td>
+        <td>
+           <input type="text" name="name" placeholder="New project">
+        </td>
+      </tr>
+      <tr>
+        <td>Client Id</td>
+        <td>
+           <input type="text" name="client_id" placeholder="Client id">
+        </td>
+      </tr>
+      <tr>
+        <td>Resources</td>
+        <td>
+          <input type="text" name="resources" placeholder="New resource">
+        </td>
+      </tr>
+      <tr>
+        <td>Description</td>
+        <td>
+          <input type="text" name="description" placeholder="New description">
+        </td>
+      </tr>
+    </table>
+    <input type="submit" value="Create Project" class="btn btn-primary">
+  </form>
+</div>
+<br>
+
+<script>
+function validateForm() {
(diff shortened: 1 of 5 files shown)


ElijahCaine:
@hsolorza This is looking really good. I have a few stretch goals for you to work on, unless @Kennric and @alxngyn think we should just go forward with this:

- Make 'Client_id' a dropdown would make this UI more usable.
- *or* Gracefully deal with a bad client ID (currently the program has a heart-attack and shows a framework error page).

Everything else looks really good though. Good job :smile:

hsolorza:
@ElijahCaine  I'm not sure how to bring the drop down button up higher so its  not so close to the footer. Any tips?

screenshot from 2016-08-30 10 53 25

Also, any preference on the color of the button?

ElijahCaine:
Hey @hsolorza I'm so sorry for not being clear in my feedback.

As far as the clients dropdown goes I meant on the `/projects/new` endpoint the client selection should be a dropdown menu and not a manual id field. Good effort on the dropdown but it looks a bit weird on the `/projects/#` page. I would get rid of that dropdown and keep it the way it was on that endpoint.

As it stands the `/projects/new` endpoint is good, the app doesn't break horribly anymore when I enter a bad id.  I don't get an error, not sure if I'm missing something there, but it does redirect me to the same page so that's a step in the right direction.

None of these changes should be too bad.
1. I really think getting rid of that dropdown is a must (sorry for the clarity error!).
2. Seeing why I'm not getting the error message when I give the DB a bad id.

Again, good work, keep it up.

views/projects/create.erb
```diff
@@ -0,0 +1,55 @@
...
+           <input type="text" name="name" placeholder="New project">
+        </td>
+      </tr>
+      <tr>
+        <td>Client Id</td>
+        <td>
+          <select  name="client_id">
+          <% for @client in @clients %>
+             <% if @client.active %>
+                <option value="<%= @client.id %>"><%= @client.id %></option>
```
alxngyn: Lets show the user the `@client.name` than ID.
So here I would change the second `@client.id` to `@client.name`

alxngyn:
+1 after the minor client ID to name view change.

ElijahCaine:
I agree with @alxngyn. Everything else is good, I like the changes and how it looks. :+1: from me

Issue #159, opened 2016-10-21
Make configuration easier


Kennric:
Allow configuration to be provied by a file or environment variables. This will make it easier to configure the app when deployed using ansible.

Kennric:
resolved bu #160

PR #160, opened 2016-10-21
Kennric/collector settings

Kennric:
fixes issue #159 

## Changes in this PR.

- [X] provide yml file based config
- [X] give env variables precedence in configuration
- [X] allow collectors to be configured consistently

## Testing this PR.

1. rake spec
2. Add an env.yml file based on env.yml-dist
3. rake spec
4. Set env variables for all config options (see env.yml-dist or environment.rb)

### Expected Output.
spec halts with an informative if there is no env.yml file nor env variables
when the file exists, spec passes
when env variables are set, spec passes
when a test mysql database is not set up (TEST_MYSQL* vars are not set), db collector spec skips with informative message

@osuosl/devs


leian7:
+1

Issue #170, opened 2016-11-04
Update docker with new env settings


hsolorza:

alxngyn:
are we going 
moving the env vars from env.config file -> docker 
or
moving the docker env vars -> env.config

hsolorza:
I believe we're moving the env vars from env.config to docker

PR #171, opened 2016-11-07
Update Env Vars - WIP

hsolorza:
fixes issue #170 

Let me know if I am missing some env vars

@osuosl/devs

### DIFF
--- a/collectors.rb
+++ b/collectors.rb
@@ -11,7 +11,7 @@ require_relative 'logging/logs'
 # systems, such as ganeti, chef, etc.
 class Collectors
   def initialize
-    @cache = Cache.new(Iam.settings.cache_file)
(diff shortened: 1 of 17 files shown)


Kennric:
I was thinking about what is actually needed in the docker env - we shouldn't need testing things, since those all get set to reasonable defaults, and the mysql test variables shouldn't be set unless we have a real mysql server somewhere to test on.

I think all we need are:

DATABASE_URL=sqlite://test.sqlite
CACHE_FILE=cachefile
LOG_FILE_PATH=logfile
GANETI_CLUSTERS=comma,delimited,list,of,clusters
DB_COLLECTOR_MYSQL_DBS=user:pass:host,user2:pass2:host2

Kennric:
+1

Issue #168, opened 2016-11-02
create db resource model


Kennric:
Model for database type resources with db name, type and server fields

Kennric:
The database table itself will be created in the 'migration' file migrations/0001_migrations.rb (base it on the existing node_resource table)

The model should be based on the NodeResource model, with the same many_to_one relationship to projects.

leian7:
Satisfied by #175

PR #175, opened 2016-11-16
add db resource model and tests, create its table

leian7:
fixes issue #168 

## Changes in this PR.

- [X] Adds database resource model
- [X] Adds tests for database resource
- [X] Creates table for database resource model

## Testing this PR.

1. ``docker-compose run --rm --service-ports dev bash``
2. ``./run-tests.sh``

### Expected Output.

Tests should pass.

@alxngyn @hsolorza @Kennric

### DIFF
--- a/migrations/0001_migrations.rb
+++ b/migrations/0001_migrations.rb
@@ -38,5 +38,16 @@ Sequel.migration do
       DateTime    :modified
       Boolean     :active, default: true
     end
+
+    create_table(:db_resources) do
+      primary_key :id
+      foreign_key :project_id, :projects
+      String      :name, unique: true
+      String      :type
+      String      :cluster
+      DateTime    :created
+      DateTime    :modified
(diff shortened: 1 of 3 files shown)


alxngyn:
++

migrations/0001_migrations.rb
```diff
@@ -38,5 +38,16 @@
       DateTime    :modified
       Boolean     :active, default: true
     end
+
+    create_table(:db_resources) do
+      primary_key :id
+      foreign_key :project_id, :projects
+      String      :name, unique: true
+      String      :type
+      String      :cluster
```
Kennric: We don't actually need a cluster field here, since dbs aren't in a cluster - but we may want to replace it with a server field to specify what server it is on (db2.o.o or mysql3.o.o, etc)

Kennric:
+1

PR #176, opened 2016-11-17
Node views

hsolorza:
fixes issue #102 

## Changes in this PR.

- [X] index
- [x] create
- [x] edit
- [x] show

## Testing this PR.

1. `docker-compose run --rm --service-ports dev bash`
2. `unicorn -l0.0.0.0:4567`
3. Create a client
4. Create a project for the client
5. Create a node for the project

@osuosl/devs

### DIFF
--- a/routes/nodes.rb
+++ b/routes/nodes.rb
@@ -63,7 +63,6 @@ module Sinatra
         params[:cluster] = nil if params[:cluster] == ''
         params[:active] = nil if params[:description] == ''
 
-
         # recieve an updated node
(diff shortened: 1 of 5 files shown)


alxngyn:
Good start. Just a few comments:

- Have Edit view include projectID so we can change it's project if needed
- Good job on Index and create view, does it's what it needs to do
- Right now the show view isn't telling us the information we want from nodes. We should add information such as plugins report values (get the values first, then maybe a graph)
download 5

Kennric:
+1

Issue #177, opened 2016-11-18
db plugin failing


Kennric:
On staging, the dbsize plugin fails with json parse errors.

PR #178, opened 2016-11-18
use correct key for cache

Kennric:
fixes issue #177 

## Changes in this PR.

- [X] Fix dbsize plugin to use correct hash key
- [X] Add some extra error catching in collectors.rb

## Testing this PR.

1. Run the app in an environment where data can be collected
2. View logs

### Expected Output.

No json parse errors.

```
$ rake spec
[...]
```

@osuosl/devs


Kennric:
@alxngyn @leian7 @hsolorza

PR #181, opened 2016-11-22
Remove db collector spec

alxngyn:
fixes issue #174 

## Changes in this PR.

- [X] Remove database collector spec tests
- [X] Remove docker configs for the testing containers

## Testing this PR.

1. Do `rake spec`.
3. See all tests pass.

### Expected Output.

```
$ rake spec
[... 100% pass ...]
```

@osuosl/devs


Kennric:
We don't want to remove the spec itself, it will skip if there isn't a db to test against. We only want to keep docker from linking the mysql container and running tests (unless it is working correctly, in which case let it go) The test only run if TEST_MYSQL_DB is set - all we have to do here is make sure that isn't set and then take the mysql container out of the docker config (or make sure the mysql tests are set up correctly)

alxngyn:
@Kennric 
Is there a way to make docker look for a flag when we run `docker-compose build` ? I can't find anything about linking vs unlinking containers on runtime.

Kennric:
I'm just closing this - the functionality we need is already in there, we should open a new issue/pr for making docker work correctly.

PR #182, opened 2016-12-06
new tables need a new migration

Kennric:
Now that the app is running a staging instance, migration for changes to the database must be in new migration files, otherwise Sequel will not add them since it knows that migration file has already been applied.

## Changes in this PR.

- [X] separate db_resources creation into a new migration

## Testing this PR.

1. migrate an existing db that was migrated before the db_resources migration was applied

### Expected Output.

db_resources table is created

@osuosl/devs

### DIFF
--- a/migrations/0001_migrations.rb
+++ b/migrations/0001_migrations.rb
@@ -39,15 +39,5 @@ Sequel.migration do
       Boolean     :active, default: true
     end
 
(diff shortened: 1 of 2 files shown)


Kennric:
@alxngyn

alxngyn:
@Kennric does this mean we are out of beta/dev and need a production branch now?

Issue #186, opened 2016-12-06
DBResource views


hsolorza:
Create views for the Databases

PR #187, opened 2016-12-06
DBResource views

hsolorza:
fixes issue #186 

## Changes in this PR.

- [X] Create
- [X] Edit
- [X] Index
- [X] Show

## Testing this PR.

1. Run the app
2. Create a db for a project

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/routes/database.rb
+++ b/routes/database.rb
@@ -1,0 +1,89 @@
+require 'sinatra/base'
+require_relative '../logging/logs'
+
+# Our app
+module Sinatra
+  # Our D Routing
+  module DBResource
+    # rubocop:disable LineLength, MethodLength, AbcSize, CyclomaticComplexity, PerceivedComplexity
+    def self.registered(app)
+      ##
+      # Database Resource
+      ##
+
+      app.get '/db/new/?' do
+        # get new database form
+        @db = DB.all
+        erb :'database/create'
+      end
+
+      app.get '/db/:id/?' do
+        # view a database
+        @db = DBResource[id: params[:id]]
+        if @db.nil?
+          MyLog.log.fatal 'routes/database: Database not found'
+          halt 404, 'database not found'
+        end
+        @projects = Project.filter(id: @db.project_id).all
+        erb :'database/show'
+      end
+
+      app.get '/db/:id/edit/?' do
+        # get database edit form
+        @db = DBResource[id: params[:id]]
+        if @db.nil?
+          MyLog.log.fatal 'routes/database: Database not found [edit]'
+          halt 404, 'database not found'
+        end
+        @projects = Project.all
+        erb :'database/edit'
+      end
+
+      app.get '/dbs/?' do
+        # get a list of all database
+        @db = DBResource.all
+        erb :'database/index'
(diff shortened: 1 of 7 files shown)


routes/database.rb
```diff
@@ -0,0 +1,89 @@
...
+        # get new database form
+        @projects = Project.all
+        erb :'database/create'
+      end
+
+      app.get '/db/:id/?' do
+        # view a database
+        @db = DBResource[id: params[:id]]
+        if @db.nil?
+          MyLog.log.fatal 'routes/database: Database not found'
```
Kennric: We might want to say "Database Resource not found" to avoid confusion with the app database

routes/database.rb
```diff
@@ -0,0 +1,89 @@
...
+        params[:name] = nil if params[:name] == ''
+        params[:type] = nil if params[:type] == ''
+        params[:server] = nil if params[:server] == ''
+        params[:active] = nil if params[:description] == ''
+
+        # recieve an updated database
+        db = DBResource[id: params[:id]]
+
+        db.update(project_id:  params[:project_id] || db.project_id,
+                  name:       params[:name] || db.name,
```
Kennric: It looks like indentation/whitespace is a little off in this block - I am surprised rubocop did not complain

routes/database.rb
```diff
@@ -0,0 +1,89 @@
...
+                  modified:   DateTime.now || db.modified,
+                  active: params[:active] || db.active)
+        redirect "/db/#{params[:id]}"
+      end
+
+      app.delete '/db/:id/?' do
+        # delete a database
+        db = DBResource[id: params[:id]]
+        db.delete unless db.nil?
+        redirect '/dbs/?' unless db.nil?
```
Kennric: isn't db nil after the delete? So won't this delete and then return 404?

Kennric: I could be wrong, please test.

views/database/edit.erb
```diff
@@ -0,0 +1,91 @@
...
+          </td>
+          <td>
+            <input type="text" name="name" placeholder="New name">
+          </td>
+        </tr>
+        <tr>
+          <td>Project Id</td>
+          <td>
+            <% if @db %>
+              <%= @db.project_id %>
```
Kennric: Whenever we display the project in a form or page, we should probably display its name, not its id, which won't mean much to people

views/database/show.erb
```diff
@@ -0,0 +1,47 @@
...
+      <thead>
+        <tr>
+          <th>Project Id</th>
+          <th>Type</th>
+          <th>Server</th>
+          <th>Active</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% for @project in @projects %>
```
Kennric: I think there should only ever be one project for a database, since the model relation is one project to many dbs, so I am not sure what this whole table is doing here

routes/database.rb
hsolorza: I haven't had rubocop complain about that

Issue #184, opened 2016-12-06
better default resource workflow


Kennric:
Right now we assign measurements for undefined resources to a default resource of that type. A better way would be to create a new resource with fqd or db name as the name, and assign to a default project belonging to the default client. Then we have many distinct resources with their own measurements, rather than one resource with thousands of conflicting measurements (because two different nodes will report different disk sizes, for example, but if both belong to one node resources, the numbers make no sense.)

PR #188, opened 2016-12-06
Kennric/better default resources

Kennric:
fixes issue #184 

## Changes in this PR.

- [X] add separate data templates for db and node caches
- [X] add methods to create resources if they don't exist already and add them to the default project
- [X] update plugins to use above methods
- [X] update spec and misc fixes to accommodate new workflow

## Testing this PR.

1. rvm all do rake spec
2. run code on staging

### Expected Output.

tests pass, data is collected properly on staging

@hsolorza @alxngyn


hsolorza:
+1

PR #189, opened 2016-12-07
Add documentation on the env vars

hsolorza:
## Changes in this PR.

- [X] Add documentation

## Testing this PR.

1. Read document
2. Understand document

@osuosl/devs


mathuin:
This may be out of scope for this particular PR, but the docs are not currently part of the build and test process.  This should change.

There's no mention in the README on the docs.  I found them in the top level directory, but they're Sphinx docs so I have to create a virtualenv by hand and install Sphinx.  When I ran make, I got fourteen errors and a warning -- all the errors were from `docs/source/meeting-notes/2016-06-30.rst` so not your fault, but the warning was from `docs/source/draft/config.rst` so it is relevant to your PR.  The warning was that this document is not included in any toctree, so that should be easy for you to fix!

The numerical list of variables is not formatted correctly.  In the initial section, the first gets `1.`, the last gets `1.`, and the rest get `#.`.  There's something in the RST or Sphinx docs that helps explain how to fix this better than I can in a comment.

They are probably better described as environment variables than ENV vars or env vars.  Read it out loud to see which sounds more natural to you.

Inside those numerical lists, use the same grammatical style -- either sentences or sentence fragments, but not both.  If you choose sentences, use punctuation.  

What is "the spec"?  From context, it appears to be the test suite.  Also, "the docker mysql" is probably better described as "the database container". 

The last section has some grammatical issues which makes it a little hard to read.  If cachefile, logfile and db_url are variables, they should get backtick'ed (or the Sphinx equivalent).  I'm not sure what exactly "it" means in this paragraph, and its meaning seems to change.  There's a typo "databse".  Also, `ENV[DB_COLLECTOR_DBS]` doesn't make much sense, especially after `DB_COLLECTOR_DBS` is mentioned in the previous sentence.  At the end, you mention `config.ru` but I couldn't find that in the code.  What does that describe?

Thank you for being patient with my unfamiliarity with the project.  I look forward to your revisions!

alxngyn (approved):

mathuin (approved):

baileysingleton (approved):
Wow these are some great docs. I think this will be very beneficial to IAM. Nice

Issue #191, opened 2016-12-08
dbsize plugin needs a migration to rename resource column


Kennric:
In the dbsize plugin, the initial migration set the resource column to node_resource instead of db_resource, a new migration is needed to fix this.

PR #192, opened 2016-12-08
migration is necessary to rename column

Kennric:
fixes issue #191 

## Changes in this PR.

- [X] Restore original initial migration
- [X] Create new migration to rename column

## Testing this PR.

1. run app connected to a database that was created with the previous migration, see that the dbsize measurement table now points to a db_resource instead of a node resource.

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs


hsolorza:
+1

PR #196, opened 2016-12-09
New Views

hsolorza:
## Changes in this PR.

- [X] Remove the Report in the Header
- [x] Client view shows their projects and their projects nodes/databases

## Testing this PR.

1. Run the app
2. Create a client
3. Create a project for that client
4. Create nodes/dbs for that project
5. View the client

### Expected Output.

screenshot from 2017-01-05 13-30-11

@osuosl/devs


routes/clients.rb
```diff
@@ -18,11 +18,33 @@ def self.registered(app)
       app.get '/clients/:id/?' do
         # view a client
         @client = Client[id: params[:id]]
+        @project = Project[id: params[:id]]
+        @node = NodeResource[id: params[:id]]
+        @db = DBResource[id: params[:id]]
```
Kennric: I don't think these are correct, you get the client's id in the params, but that isn't the same as the id of the project that belongs to that client. 

With reciprocal relationships, you should be able to just fetch the client, and then refer to its projects like @client.projects and the projects resources like project.node_resources..

see http://sequel.jeremyevans.net/rdoc/files/doc/association_basics_rdoc.html - especially Methods Added and Filtering by Associations sections

routes/clients.rb
```diff
@@ -18,11 +18,33 @@ def self.registered(app)
...
+
         if @client.nil?
           MyLog.log.fatal 'routes/clients: Client not found'
           halt 404, 'Client not found'
         end
-        @projects = Project.filter(client_id: @client.id)
+
+        if @project.nil?
```
Kennric: I don't think you need this if/else - just fetch the projects from this client, (client.projects)  and if it's an empty dataset, you can use that as a condition for the if in the template (<% if @projects %>Stuff...)

similarly with node_resources and db_resources

views/clients/show.erb
```diff
@@ -1,47 +1,122 @@
...
-        <h2>Client: <%= @client.name %> -
-        <a href="/clients/<%= @client.id %>/edit">edit</a>
-        </h2>
-        <p>
-          Name: <%= @client.contact_name %> <br>
-          Email: <%= @client.contact_email %> <br>
-          Description: <%= @client.description %> <br>
-          Active: <%= @client.active %>
-        </p>
+<% unless @pro_nil  %>
```
Kennric: I think you can just say <% if @projects %> here

Issue #197, opened 2016-12-09
dbsize measurmeent table value column is too small


Kennric:
Since we are measuring db size in bytes, some of the numbers overflow the postgres int type. Migrate the measurement table to use bigint instead

PR #198, opened 2016-12-09
dbsize is a big int

Kennric:
fixes issue #197 

## Changes in this PR.

- [X] Plugin migration to change the value column for db_size_measurements to type bigint

## Testing this PR.

1. Examine an existing app database
2. Run the app or manually run this migration

### Expected Output.

value column is now type bigint

@osuosl/devs


alxngyn:
++

Issue #201, opened 2016-12-09
project view should have more info


alxngyn:
Project view needs more info

- remove client table and just put linking info in project under header
- add table of nodes resources
- add table of db resources

Kennric:
I think this is basically implemented, file more specific issue for improvements.

PR #203, opened 2016-12-10
Update project view

alxngyn:
fixes issue #201

## Changes in this PR.

- [X] update project views
- [X] remove clients tbale and add nodes and DBs table
- [X] remove footer
- [X] remove resources field and related spec test
- [X] removed Node and DB ID and replaced it with table headers

## Testing this PR.

1. Do run app with scheduler to fill in data.
2. Run docker with web.
3. See projects page.

### Expected Output.

screencapture-pink-workstation-cass-oregonstate-edu-4567-projects-7-1481921706812

@osuosl/devs


routes/projects.rb
```diff
@@ -25,6 +25,21 @@ def self.registered(app)
           halt 404, 'Project not found'
         end
         @clients = Client.filter(id: @project.client_id).all
+        if @clients.nil?
+          MyLog.log.fatal "routes/projects: Project's clients not found"
+          halt 404, "Project's Client not found"
+        end
+        @nodes = NodeResource.filter(project_id: @project.id).all
```
hsolorza: I don't think that the nodes and dbs should have an error if they aren't present, a project doesn't have to have either

Kennric:
Looking good - a couple of nitpicks - the 'Resources:' line at the top isn't useful - I think we're going to deprecate that field on the model.

Some table headers would be good for each resource type, and be sure the table column headers correspond to the actual resource fields. I don't think the id field is very useful in this view.

routes/projects.rb
alxngyn: you're right, ill make sure it checks in the view instead

alxngyn:
@hsolorza @Kennric 
updated views based on feedback and removed node/db checks in routing.
Screenshot has also updated.

routes/projects.rb
```diff
@@ -25,6 +25,12 @@ def self.registered(app)
...
+        @dbs = DBResource.filter(project_id: @project.id).all
```
Kennric: If you use reciprocal associations in the model, you should be able to just get the client, nodes and dbs from the project object: project.nodes, project.client
See http://sequel.jeremyevans.net/rdoc/files/doc/association_basics_rdoc.html and the kennric/reciprocal_relations branch

views/projects/show.erb
```diff
@@ -3,43 +3,90 @@
...
+        <a href="/clients/<%= @client.id %>" >
+          <%= @client.name %>
+        </a>
+      <% end %>
+    <% end %>
+
+    <br>
     Description: <%= @project.description %> <br>
-    Active: <%= @project.active %><br>
+    Active State: <%= @project.active %><br>
```
Kennric: It's a minor thing, but you could do something like
```
<%= @project.active ? "Active" : "Inactive" %>
```
To either print Active or Inactive depending on the field value. You could even have it print a span around the text with a green or red class to make it more obvious.

routes/projects.rb
alxngyn: I prepped my branch to use it once that branch goes into develop

views/projects/show.erb
```diff
@@ -3,43 +3,86 @@
...
+          <%= @client.name %>
+        </a>
+      <% end %>
+    <% end %>
+
+    <br>
     Description: <%= @project.description %> <br>
-    Active: <%= @project.active %><br>
-  </p>
+    Active State: <%= @project.active ? "Active" : "Inactive" %>
```
Kennric: I think "Active State:" is redundant, just print whether it is active or not.

views/projects/show.erb
```diff
@@ -3,43 +3,86 @@
     <a href="/projects/<%= @project.id %>/edit">edit</a>
   </h2>
   <p>
-    Client Id: <%= @project.client_id %> <br>
-    Resources: <%= @project.resources %> <br>
+    <% for @client in @clients %>
```
Kennric: We really should only ever have one client per project, so we don't need a full loop here, just take the first in the clients array.

views/projects/show.erb
```diff
@@ -3,43 +3,86 @@
...
+              </a>
+            </td>
+            <td>
+              <%= @node.type %>
+            </td>
+            <td>
+              <%= @node.cluster %>
+            </td>
+            <td>
+              <%= @node.active %>
```
Kennric: Change this to the same construct used to display Active or Inactive in the client block above, and change this column header to say "Status"

Issue #183, opened 2016-12-06
Visuals for Node Information


hsolorza:
When viewing a node, there should be graphs that visualize the node's data.

PR #205, opened 2016-12-20
Plugin data for node view

alxngyn:
fixes issue #183

## Changes in this PR.

- [X] Edited views to take average of the available plugin data from DB

## Testing this PR.

1. Do rake.
2. See table headers and tables

### Expected Output.

Example with generated random nodes with random values for DiskSize:
screencapture-pink-workstation-cass-oregonstate-edu-4567-node-2-1482269871167

@osuosl/devs

### DIFF
--- a/views/nodes/show.erb
+++ b/views/nodes/show.erb
@@ -11,7 +11,7 @@
   <table class="table">
       <thead>
         <tr>
-          <th>Project Id</th>
+          <th>Name</th>
           <th>Type</th>
           <th>Cluster</th>
           <th>Active</th>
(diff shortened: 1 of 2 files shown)


views/nodes/show.erb
```diff
@@ -3,25 +3,24 @@
     <a href="/node/<%= @node.id %>/edit">edit</a>
   </h2>
   <p>
-    Project Id: <%= @node.project_id %> <br>
```
hsolorza: I think that the Project Id is something that should still be displayed in the views

alxngyn: While i think it's useful for debugging and url forwarding I don't think it's useful for the users.

Kennric:
Also, can you add a last-updated date on the measurements table?

alxngyn:
Updated code from feedback
download 7

views/nodes/show.erb
Kennric: Make this the project name and remove the projects table below. A node only belongs to one project, and we don't need that project's details here - just a name that links back to the project view.

Issue #211, opened 2017-01-11
edit link for dbs produces an error


Kennric:
See  http://iam.osuosl.org/dbs

click on the edit button for any database

```
application error
NoMethodError at /db/1/edit
undefined method `name' for nil:NilClass

    file: edit.erb location: block in singleton class line: 32 
```

hsolorza:
fixed with #215

PR #215, opened 2017-01-11
Fix database edit

hsolorza:
fixes issue #211 

## Changes in this PR.

- [X] Fix db edit

## Testing this PR.

1. start Iam 
2. view the db edit

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/routes/database.rb
+++ b/routes/database.rb
@@ -35,6 +35,7 @@ module Sinatra
           MyLog.log.fatal 'routes/database: Database Resource not found [edit]'
           halt 404, 'database resource not found'
         end
+        @project = Project.filter(id: @db.project_id).all.first
         @projects = Project.all
         erb :'database/edit'
       end
--- a/routes/database.rb
+++ b/routes/database.rb
@@ -13,10 +13,7 @@ module Sinatra
 
       app.get '/db/new/?' do
         # get new database form
-
         @projects = Project.all
-
-        puts @projects
         erb :'database/create'
       end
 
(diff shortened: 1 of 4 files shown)


routes/database.rb
```diff
@@ -35,6 +35,7 @@ def self.registered(app)
           MyLog.log.fatal 'routes/database: Database Resource not found [edit]'
           halt 404, 'database resource not found'
         end
+        @project = Project.filter(id: @db.project_id).all.first
         @projects = Project.all
```
Kennric: it looks  like `@projects` being used to populate the project pull-down list.. it feels a bit redundant to query for one project and also query for all of them. I think you can just get the specific project out of `@projects` - that's a Sequel dataset, so you can filter it without doing a new query. So get Projects.all, then get this specific project out of that dataset like `@projects.where(:id => @db.project_id)`

views/database/edit.erb
```diff
@@ -26,7 +25,7 @@
           </td>
         </tr>
         <tr>
-          <td>Project Id</td>
+          <td>Project Name</td>
```
Kennric: You can just say 'Project' here

routes/database.rb
alxngyn: i agree
It should be `@projects = Project.filter(id: @db.project_id).all` (without the first)
then in the view, it should iterate through the projects

views/database/edit.erb
```diff
@@ -26,7 +25,7 @@
...
           <td>
             <% if @db %>
```
alxngyn: If a DB exists without a parent project
this line is where the view is trying to parse a nil object

should change to something like `<% if @project %>`

Issue #220, opened 2017-01-13
Client summary


Kennric:
On the client view page, add a table to the top that summarizes all of the billable numbers for that client. Currently these are RamSize, VCPU, and DiskSize for nodes, and the total size in GB (minimum 1GB) for databases.

If the DiskTemplate measurement is 'drbd' on a node, all of the numbers for that node must be multiplied by 2.

```
Summary
--------------------------------------------------------------------------------------------
  Nodes            Ram (GB)            VCPU            Disk Size (GB)            Project
  ------------------------------------------------------------------------------------------
  node x            4                  2                10                       Project A
  node y            2                  1                5                        Project A

  ------------------------------------------------------------------------------------------
  Databases            Size (GB)             Project
  ------------------------------------------------------------------------------------------
  database x            10                  Project A
  database y            1                   Project B
```

alxngyn:
shouldn't we just have a column for DBRD? I think multiplying the data by 2 makes it really confusing versus multiplying the rate by 2

Kennric:
@alxngyn multiplying the measurements is the requested feature, so that's what we're doing. We don't want to multiply -and- indicate drbd, as that could confuse the viewer, who may not know whether the numbers are already multiplied or whether they need to do it. When we have rates in the system, we can change this and do the math there, but for now, we are displaying what will be entered into the invoice.

Kennric:
Done, as described in this issue, will create new issues for adjustments.

alxngyn:
resolved by #222

PR #222, opened 2017-01-13
Resource Summary

hsolorza:
fixes issue #220 

## Changes in this PR.

- [X] Tables for resource summary
- [X] Fixed layout of projects table
- [x] DiskTemplate 'drbd' multiplies node # by 2

## Testing this PR.

1. Run app
2. View a client that has projects and resources
3. See summary table

### Expected Output.

**Full client page:**

fullsize

@osuosl/devs

### DIFF
--- a/views/clients/show.erb
+++ b/views/clients/show.erb
@@ -97,91 +97,103 @@
       </td>
     </tr>
    </table>
+
+
+<!-- ooooooooooooo -->
    <table class="table">
-     <h4> Projects </h4>
-       <% for @project in @projects %>
-           <thead>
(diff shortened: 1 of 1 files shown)


views/clients/show.erb
```diff
@@ -8,92 +8,195 @@
     Description: <%= @client.description %> <br>
     <%= @client.active ? 'Active' : 'Inactive' %>
 </p>
+
 <% unless @projects.nil? %>
```
alxngyn: should print something if projects are nil instead of letting the user see blank

Issue #221, opened 2017-01-13
Collection Status model


Kennric:
It would be useful for several things to have a record of collector runs, and perhaps other stats.

Each model object would consist of a collector name, timestamp, perhaps duration (the collector method should be able to measure run time), maybe total rows in the collected data, a status (failed, succeeded, succeeded with some errors, etc)and anything else that might be useful to capture about collector runs.

alxngyn:
should the model parse the logs or should the collector write a row to the db?
Maybe we can add a feature that parses the logfile in the view and bypass the model?

Kennric:
The collector should write a row to the table on each run, there is no reason to write out to a file and then read back in.

PR #223, opened 2017-01-18
Collect stats model

alxngyn:
fixes issue #221

## Changes in this PR.

- [X] Change models.rb
- [X] Add migrations file
- [x] Create model tests

## Testing this PR.

1. Do `rake spec`. 
2. See test pass.

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/migrations/0003_migrations.rb
+++ b/migrations/0003_migrations.rb
@@ -1,0 +1,14 @@
+Sequel.migration do
+  transaction
+  change do
+    create_table(:collector_stats) do
+      primary_key :id
+      String      :name
+      DateTime    :created
+      Time        :start
+      Time        :end
+      Integer     :duration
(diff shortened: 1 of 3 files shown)


migrations/0003_migrations.rb
```diff
@@ -0,0 +1,14 @@
+Sequel.migration do
+  transaction
+  change do
+    create_table(:collector_stats) do
+      primary_key :id
+      String      :name
+      DateTime    :created
+      Time        :start
+      Time        :end
+      Integer     :duration
```
Kennric: If we have start and end times, duration is redundant

alxngyn: you're right, i'll remove it

Issue #225, opened 2017-01-19
Move data query into routes code


Kennric:
for views that show measurement data, we should query/calculate the relevant information in the route code (or abstract  into util) and send a data object to the view, rather than querying and calculating in the view code itself. The structure of the data view should allow easily iteration in a table to display all the relevant data columns.

PR #229, opened 2017-01-20
Kennric/summary fixes

Kennric:
fixes issue #225 

## Changes in this PR.

- [X] Creates plugin matrix and project report methods in a Report class in util

## Testing this PR.

1. call the project report method with the name of a project 

### Expected Output.

receive a hash of its resources and their measurements

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/lib/BasePlugin/plugin.rb
+++ b/lib/BasePlugin/plugin.rb
@@ -89,7 +89,7 @@ class TestingPlugin < BasePlugin
   #  super()
   def initialize
     name = 'TestingPlugin'
-    resource_name = 'resource'
+    resource_name = 'node'
     units = 'units'
     table = :test_plugin_measurements
     db_column = :test_plugin_ver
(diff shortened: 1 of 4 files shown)


Kennric:
@alxngyn, @hsolorza

hsolorza:
When I try to view a client that has a project I get this error:
undefined method `resource_resources' for #<Project:0x007f8c01f78598> Did you mean? resources=
     file: util.rb location: block in project_data line: 150

alxngyn:
getting the same client error as @hsolorza 
ss of the backtrace
fireshot capture 23 - nomethoderror at _clients_1 - http___localhost_4567_clients_1

Kennric:
This is caused by the existence of the Testing Plugin, which the base plugin spec uses for testing the base plugin methods. The testing plugin declares that it measures resources of type "resource". Since we now explicitly relate projects to resource types, but have not created a relationship from projects to "resource" resources, this breaks when the report code tries to find a project's "resource" resources  by calling projects.resource_resources. 

This was not breaking on staging when I was testing, because I changed the testing plugin's resource type to node on staging (this breaks spec on docker, because the testing plugin migration still creates a 'resource' column in its measurement table, note a 'node' one.

The solution, I think, is simply to skip the testing plugin in the code that gathers all the plugins and their available measurements.

hsolorza:
+1

alxngyn:
++

Issue #224, opened 2017-01-19
Change client and project views to summary style


Kennric:
Based on feedback from @ramereth:

Change the current client view to be more like the recently created Summary table - list all the nodes and databases along with their measurements, and leave the resource details (cluster, type, etc) to the individual resource view. This view will be essentially the same as the Project view, except that a client view may contain data from multiple projects.

Create a new view, 'billing' which will display a condensed summary table that has only the totals for all VPCU, RAM, DB size, etc. In this view, ganeti node measurements will be multiplied by 2 if the disk template is drbd. This view will also contain a date range selector for adjusting the date range and quick links to select month to date, last calendar month, and last quarter.

Issue #238, opened 2017-02-10
Inconsistent table colors


hsolorza:
With the new bootstrap style, all the tables in IaM should be grey. Though when viewing a client, the tables are white.
![screenshot from 2017-02-10 14 31 29](https://cloud.githubusercontent.com/assets/20174370/22846771/00a1d774-ef9e-11e6-8ed7-a7ce3d1b21e7.png)

alxngyn:
@hsolorza is it just clients? I don't remember adding any client specific CSS

hsolorza:
@alxngyn this seems to only occur with Client and Node.

hsolorza:
@alxngyn  yes, it seems to only occur with the Clients

Issue #236, opened 2017-02-01
No graceful error for duplicate names


hsolorza:
If a client/project/node/database is being created and is given the same name as something that already exists, a sequel error is thrown as 'the column name is not unique'. In the event of a duplicate name, there should be something that catches the error and says something like 'that name is already in use'.

PR #241, opened 2017-03-03
Client billing

hsolorza:
fixes issue #224  #238 #236 #244 

## Changes in this PR.

- [X] Client tables generated using project_data hash
- [x] Create billing page
- [X] Inconsistent table colors fixed
- [x] Adjust time frame for billing
- [x] Fix duplicate name errors 
- [x] Fix Fixnum error when viewing a client

## Testing this PR.

1. View a client 
- View the billing page
- Adjust the time frames for billing

2. Create a new client with a name that already exists
- Receive the duplicate name error

### Expected Output.

The error displayed when a duplicate client is created:
screenshot from 2017-03-13 14 22 59

The billing page:
screenshot from 2017-05-08 14-44-18

```
$ rake spec
[...]
```

@osuosl/devs


lib/util.rb
```diff
@@ -175,30 +180,41 @@ def self.project_data(project)
...
+                     sum[res_type].store(key, value)
+                   end# if sum
+                 end #if
+               end #unless
+             end #data
+           end #res_hash
+         end #resource_hash
+        end #resource
+      end #project_resource
+    end #input_hash
```
ElijahCaine: gross
Oh gawd. This looks very not fun.

hsolorza: The hash that I'm trying to cycle through doesn't make attempting to sum up all the measurements easy :/  There probably is an easier way to do this, I just haven't found it yet...

ramereth:
@hsolorza I'm still working on getting my dev environment working however does this branch include totals at the bottom of the VM view and Database view?

hsolorza:
@ramereth Im not sure what you mean. In this branch the only place where there are totals is in the Clients Billing view.

lib/util.rb
```diff
@@ -211,4 +221,51 @@ def self.project_data(project)
     end
     project_data
   end
+
+  # this method takes a hash of data and two dates. It takes the data that falls
+  # between the two dates, then returns the sum of their measurements
+  def self.data_in_date_range(input_hash, start_date, end_date)
```
Kennric: I think we're doing unnecessary work here - when we fetch the data from the plugin's report method , we can send a date range, so we only ever fetch the data in that range rather than fetching it all and pulling the relevant range in a new method. When the data is fetched for each project's resource's measurement plugin (project_data method in Report class, line 209 util.rb), we should call the report method with the given date range.

To do that, add startdate, enddate parameters to the project_data method (default nil), and call project_data from the route with the dates you receive from the form.

Then this method only needs to sum up the given data, since the data should only contain the correct time range.

alxngyn:
I'm getting:

> TypeError at /clients/2/billing/1492473600/1492992000
> start_time and end_time should be Time objects

This is what I did for date input:
screenshot from 2017-04-21 15 37 35

alxngyn:
+1 works for me!
I'd say make an issue for the bug that the dates disappear on load

ramereth:
I noticed the units are off when displaying them. Is this related to #246?

ramereth:
I find the current layout for the billing view to be not ideal. I think it should be formatted more closely to how our invoices look. Something like the following would be much better:

```
For time period 04/01/2017 - 04/30/2017
Item            Description        Qty
----------------------------------------
Database        foo_db              2
Database        bar_db              5
                  Database Subtotal 7
----------------------------------------
CPU             foo.example.org     2
CPU             bar.example.org     6
                       CPU Subtotal 8
----------------------------------------
```

Then later on, we could add another column which may list the price and then the total each line item. This allows me and Pam to see a detailed list of where the costs are and the total which we would put in the actual invoice.

hsolorza:
@ramereth  Is the Database your referring to the plugin DBSize?

ramereth:
@hsolorza yes

ramereth:
@hsolorza this looks great overall but I did run into one bug. When I tried to select dates outside of the range I have data for I got an error.

ramereth:
@hsolorza to provide some context on how to replicate the problem, I imported some production data using @Kennric branch in #254 first, and then switched to yours and ran ``rake migrate`` and then brought the instance up. When I tried to select a date range for the Drupal client from May 1 to May 30 I got the following error [output](https://github.com/osuosl/iam/files/1029547/ArgumentError.at._clients_1_billing_1493596800_1496188800.pdf).

I'm not sure if it's an artifact from his branch or yours but I was surprised when that happened. If I try to use the date range with default data imported I don't get this error.

Issue #242, opened 2017-03-16
Delete links


Kennric:
Objects created within the app (projects and clients) may sometimes need to be deleted. Add a delete link in addition to Edit links.

PR #245, opened 2017-04-06
Delete records

hsolorza:
fixes issue #242 

## Changes in this PR.

- [X] Delete client/project when set to inactive
- [X] nodes/databases get owned by default
- [x] Tests

## Testing this PR.

1. Create a client/ project with resources
2. Set client/project to inactive
3. The links will be deleted

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs


Kennric:
I think what we want to do is not delete the node and db resources, but dis-associate them from the project. These are automatically created by the collector code, so we shouldn't manually delete them, but change the project they point to to 'default'.

We could also re-assign projects orphaned by a client delete to the default client, but I don't think that's very useful.

So, delete the project, but if it has resources (we should also dynamially determine what kind of resources it has so we don't have to code node_ and db_) then modify those resources to point their project_id to the default project.

Also, it occurs to me that we -really- don't want to delete the default client or project, so we should do a  check to prevent that.

Finally, don't forget to write tests.

lib/util.rb
```diff
@@ -122,6 +122,17 @@ def self.days_in_range(data)
 
 # methods for gathering measurement data into hashes
 class Report
+  # this method returns an array of all the resources (without their
+  # measurements) a project has associated with it
+  def self.get_resources(project)
+    all_resources = []
+    plugin_matrix.each do |resource_type, _measurements|
+      all_resources.push(project.send("#{resource_type}_resources"))
```
Kennric: What happens here if the project doesn't have db resources, but db is in the plugin matrix?

routes/clients.rb
```diff
@@ -80,15 +80,25 @@ def self.registered(app)
                       contact_email: params[:contact_email] || client.contact_email,
                       contact_name: params[:contact_name] || client.contact_name,
                       active: params[:active] || client.active)
-        redirect "/clients/#{params[:id]}"
-      end
 
-      app.delete '/clients/:id/?' do
```
Kennric: Did you mean to remove the delete route? This code seems to live under app.patch, which is used for editing a client, but probably shouldn't be used to delete one.

routes/clients.rb
```diff
@@ -80,15 +80,25 @@ def self.registered(app)
...
-        404
+        # if the active state is false, all projects are deleted and resources
+        # are re-assigned to the default project.
+        unless client.active || client.name == 'default'
+          unless client.projects.empty?
+            projects = client.projects
+            projects.each do |project|
+              resources = Report.get_resources(project)
+              unless resources.empty?
+                resources.each { |resource_type| resource_type.update(project_id: 1) }
```
Kennric: This is correct, but it would be safer to look up the default project's id by name instead of hard-coding 1 (I think in current production database, the id is actually 9)

routes/database.rb
```diff
@@ -104,15 +104,13 @@ def self.registered(app)
                   server:    params[:server] || db.server,
                   modified:   DateTime.now || db.modified,
                   active: params[:active] || db.active)
-        redirect "/db/#{params[:id]}"
-      end
 
-      app.delete '/db/:id/?' do
```
Kennric: as with clients, I think we want an actual delete method here, but in this context (calling delete on a database), we don't need to re-assign it, we just want to deactivate it. Re-assigning is only needed if we've deleted the parent project. Same comment regarding nodes.

routes/projects.rb
```diff
@@ -70,15 +70,21 @@ def self.registered(app)
         project.update(name:  params[:name] || project.name,
                        description: params[:description] || project.description,
                        active: params[:active] || project.active)
-        redirect "/projects/#{params[:id]}"
-      end
 
-      app.delete '/projects/:id/?' do
```
Kennric: Same with clients above, lets have an actual delete method, not patch, and look up the default id by name.

routes/clients.rb
```diff
@@ -80,15 +80,25 @@ def self.registered(app)
...
-        redirect '/clients' unless client.nil?
-        404
+        # if the active state is false, all projects are deleted and resources
+        # are re-assigned to the default project.
+        unless client.active || client.name == 'default'
+          unless client.projects.empty?
+            projects = client.projects
+            projects.each do |project|
+              resources = Report.get_resources(project)
+              unless resources.empty?
```
Kennric: Ideally, both the client delete and project delete could call the same code re-assigning all the resources for a specific project. Not sure it's worth breaking out into util at this point, but something to think about.

routes/clients.rb
hsolorza: I discovered that the app.delete was never being called. There is a way to call the delete method, https://gist.github.com/victorwhy/45bb5637cd3e7e879ace but I was not successful in getting the Post and Delete to work simultaneously on the same form.

I thought it would be easier just to have the app.patch check if the active state was changed, then either keep or delete the resource accordingly.

routes/clients.rb
```diff
@@ -80,13 +80,25 @@ def self.registered(app)
...
+
         redirect "/clients/#{params[:id]}"
       end
 
       app.delete '/clients/:id/?' do
         # delete a client
         client = Client[id: params[:id]]
-        client.delete unless client.nil?
+        unless client.name == 'default'
+          unless client.projects.empty?
```
Kennric: This code could be a model method, `client.remove`, which would check if it owns projects and use a project model method `project.remove` int he project model, which would reassign resources using a `project.reassign_resources` method, then call `project.delete`. Once this is complete, the `client.remove` method would call `client.delete`.

alxngyn:
I'm getting this foreign key constraint:
screenshot from 2017-04-26 14 58 56

Steps to reproduce:
1. create client
2. Created project attached to client
3. created node attached to project
4. created db attached to project
5. Go back to client list -> details -> edit -> Delete button -> Error (pictured)

Helpful details:
```
sinatra.error	
#<Sequel::ForeignKeyConstraintViolation: SQLite3::ConstraintException: foreign key constraint failed>
sinatra.route	
DELETE /clients/:id/?
```

Kennric:
@alxngyn I can't replicate that error with either imported test data or by adding clients/nodes/databases.

models.rb
```diff
@@ -15,6 +15,16 @@ def validate
     super
     errors.add(:name, 'cannot be empty') if !name || name.empty?
   end
+
+  def remove(client)
```
Kennric: I don't think you need to take client as a parameter here, since this is an instance method that will be called like `client.delete` - the client instance is already there, there is no reason to send it a copy of itself.

models.rb
```diff
@@ -15,6 +15,16 @@ def validate
...
+    unless client.projects.empty?
```
Kennric: I think you can refer to the current instance as `self` like in Django models, so you could say `unless self.projects.empty?`

models.rb
```diff
@@ -31,6 +41,18 @@ def validate
     super
     errors.add(:name, 'cannot be empty') if !name || name.empty?
   end
+
+  def reassign_resources(project)
```
Kennric: As with client, I don't think you need the project parameter

routes/projects.rb
```diff
@@ -70,13 +70,16 @@ def self.registered(app)
...
         redirect "/projects/#{params[:id]}"
       end
 
       app.delete '/projects/:id/?' do
         # delete a project
         project = Project[id: params[:id]]
-        project.delete unless project.nil?
+        # disassociate this projects' resources to the default project and
+        # delete this project
+        project.reassign_resources(project) unless project.name == 'default'
```
Kennric: In addition, we should probably not show a delete button on the default client/project, it'll be confusing that it is there but doesn't do anything.

views/database/edit.erb
```diff
@@ -70,21 +70,18 @@
...
-            <% end %>
+            <input type="submit" value="Update" class="btn btn-primary">
           </td>
           <td>
-            <input type="radio" name="active" value="True"> True
-            <input type="radio" name="active" value="False"> False
+          </form>
+          <form action="/dbs/<%= @db.id %>" method='post'>
+            <input type="hidden" name="_method" value="delete">
+            <input type="submit" value="Inactive" class="btn btn-primary">
```
Kennric: Very small nitpic - I think the button should be labeled "Deactivate" or "Delete"

PR #257, opened 2017-04-28
Hsolorza/join models

hsolorza:
fixes issue #

## Changes in this PR.

- [x] Add node_resources_project model and migrations
- [X] Add db_resources_project model and migrations

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/migrations/0002_migrations.rb
+++ b/migrations/0002_migrations.rb
@@ -4,7 +4,6 @@ Sequel.migration do
     create_table(:db_resources) do
       primary_key :id
       foreign_key :project_id, :projects
-      foreign_key :db_project_id, :db_resources_projects
       String      :name, unique: true
       String      :type
       String      :server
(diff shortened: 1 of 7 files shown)


migrations/0002_migrations.rb
```diff
@@ -4,6 +4,7 @@
     create_table(:db_resources) do
       primary_key :id
       foreign_key :project_id, :projects
+      foreign_key :db_project_id, :db_resources_projects
```
alxngyn: I don't think `node` or `DB` resources need to know they are in a `resource_projects` table
For example, If we are referencing the resource<->projects relationships I don't think we will look a the `db_resources` table for them

alxngyn:
Looks good. +1
Side note:  Is the plan for the names to be on the SKU model or on the pairs?

PR #260, opened 2017-05-10
Chef collector

DevonCash:
addresses issue #166 

## Changes in this PR.

- [X] Created a test for a chef collector

## Testing this PR.

1. Run rake spec

### Expected Output.

```
$ rake spec
[...]
```
Test 'IaM Chef Collector' should fail to correctly read and store data
@osuosl/devs


spec/collectors/chef_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    server.stop
+  end
+
+  it 'Correctly reads and stores data' do
+    c = Collectors.new
+    # call chef collector
+    c.collect_chef(
+      server.url,
+      'test',  # an arbitrary user name
+      ENV['TEST_CHEF_PEM'] # path to a properly formatted private key
```
Kennric: Are we expected to set this before running tests?

Kennric (commented):
Can you document a little more how the tests will interact with the chef zero server? Are there variables or settings we need to set up before running tests?

Kennric (approved):

alxngyn:
- lets get the collector combined into this PR so we don't merge in a broken Travis build
- Also, looks like you edited the dataflow image but I only see the dimensions were edited. Do we need that change?
- Other than that, looks good

panntomime:
Collector function is written and tests are passing, all that's left now is hooking it into the scheduler and setting it up with real chef credentials. 
@Kennric what's my next step?

collectors.rb
```diff
@@ -57,6 +57,34 @@ def collect_ganeti(cluster)
...
+        node_name = n.name || 'unknown'
+        ram_total = n.automatic['memory']['total'] || 'unknown'
+        ram_free = n.automatic['memory']['free'] || 'unknown'
+        cpus_total = n.automatic['cpu']['total'] || 'unknown'
+        cpus_real = n.automatic['cpu']['real'] || 'unknown'
+
+        # Collect disk size
+        disk_size = disk_usage = disk_count = 0
+        n.automatic['filesystem2']['by_device'].each do |_key, device|
+          next unless device['fs_type'] == 'ext4'
```
Kennric: Are there other filesystems that might be worth counting? Do you see ext3 or others in the raw data?

collectors.rb
```diff
@@ -57,6 +57,34 @@ def collect_ganeti(cluster)
     @node_cache.write
   end
 
+  def collect_chef(url, client, key)
+    chef_template = ERB.new File.new('cache_templates/chef_template.erb').read,
+                            nil, '%'
+
+    ChefAPI::Connection.new(endpoint: url, client: client, key: key) do |connection|
+      connection.nodes.each do |n|
+        node_name = n.name || 'unknown'
```
Kennric: Are we going to have collisions if multiple nodes don't have a name? Do we know if it is even possible for nodes not to have a name?

alxngyn (approved):
Looks good!

Kennric (commented):
Looks good, just have a couple of questions

collectors.rb
panntomime: There's also a set marked temp, but I'm only looking at the data from one node so there may be others

collectors.rb
panntomime: I don't think it's possible, but I usually err on the side of caution. If you'd prefer to throw an error in the event of an unnamed node, we could do that too

collectors.rb
Kennric: ext3/4 are the most likely to have relevant data, let's check those two, unless someone from infra says there are others.

collectors.rb
Kennric: I think this is good.

collectors.rb
Kennric: looks like XFS is also in use, so add that to the list. It is probably a good idea to split these out so that we can distinguish the filesystems for billing, just in case. So a line per filesystem type, and the plugin for now can just add those up

hsolorza:
I receive this error when testing this PR:
screenshot from 2017-08-16 14-31-41

Issue #266, opened 2017-05-25
DRBD/Plain indicator for Billing view


ramereth:
It would be nice to display whether a specific item listed in the billing view is counted for drbd or plain. This is only for VM related items (disk, ram, vcpu). I'm not sure what the best way to show it visually (perhaps an asterisk?).

Issue #267, opened 2017-05-25
Display units in billing view


ramereth:
It would be great to display the units shown in the billing view.

Issue #265, opened 2017-05-25
Database billing accounting logic


ramereth:
Something I neglected to mention regarding how we bill for databases is the fact that we bill for a minimum of 1G per database. The current billing view shows the actual disk total (which still might be good to show regardless), but it should also include the Billing total using the logic I mentioned.

So for example, if we have the following totals:

```
Database answ_ros_askbot1 1.7
Database ros_hotcrp2013 0.0
Database ros_mt 0.06
Database ros_org_wp 0.0
Database ros_wordpress 0.01
Database Total: 1.77
```

According to our billing method, this should total out to be **5.7**.

PR #268, opened 2017-06-02
Billing units

hsolorza:
fixes issue #266 #265 #267 

## Changes in this PR.

- [X] Units now present
- [X] Fix billing logic for Database
- [X] DRBD indicator

## Testing this PR.

1. View current month
2.  There are no measurements for current  month so plugins show N/A.

3. View month range from Nov - June
4.  Database values, if less than 1, should show 1 (actual value).

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/lib/util.rb
+++ b/lib/util.rb
@@ -124,6 +124,7 @@ class DataUtil
   # the type of data, then adds their value to the final hash of sums
   def self.sum_data(sums, key, value, drdb)
     drdb = 1 if drdb.nil?
+    # convert value to minimum billable value if < 1
     value = value < 1 ? 1 : value
     # if sums already contains this key, add the value to the existing value;
     # else add the key and value to sums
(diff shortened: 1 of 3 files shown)


Kennric:
@ramereth - question: if no data is found in the database for a time range, should that round to 1 as well? No data would seem to indicate that the host is no longer extant.

ramereth:
@Kennric no, it should probably display ``N/A``

ramereth:
@hsolorza I like how you put the actual size in parens for the database information. I wonder if we should do the same for hosts that have DRBD. Show the actual size in parens and the size to the left is the calculated size for billing.

ramereth:
@hsolorza I think having some kind of description about what the values in ``()`` mean would be good. Something like _Values between () are actual values compared to billable values which vary on the fee type. Please see fee book for more information_.

Issue #269, opened 2017-06-06
Resource totals on Client Page


ramereth:
We currently show values for each resource a client uses. You can see totals in the billing view, however the math there might be different than actual numbers. I would be preferred to put totals at the bottom of each resource shown.

![screenshot from 2017-06-06 15-26-07](https://user-images.githubusercontent.com/48997/26854805-9c6b9cf6-4acc-11e7-829a-19f6f9e34cbe.png)

For example, Under the "Node" section it would be great to see totals for disk sizes, ram size, etc at the bottom of each section and then the same for the DB section.

PR #273, opened 2017-06-08
Client resource totals

hsolorza:
fixes issue #269 

## Changes in this PR.

- [X] Totals display on client show page
- [X] Minor updates to the show page
- [X] Update the edit views

## Testing this PR.

1. View a client
2. See totals for each resource

### Expected Output.

screenshot from 2017-06-23 13-02-07

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/views/clients/show.erb
+++ b/views/clients/show.erb
@@ -1,5 +1,5 @@
 <div class="container">
-<h2> Client: <%= @client.name %>
+<h2> Client: <%= @client.name %> -
 <a href="/clients/<%= @client.id %>/edit"> Edit</a>
 </h2>
 <p>
@@ -9,6 +9,7 @@
     <%= @client.active ? 'Active' : 'Inactive' %><br>
     <a href="/clients/<%= @client.id %>/billing"> <h4>Billing</h4> </a>
 </p>
+<br>
 
 <% if @projects.empty? %>
   <p>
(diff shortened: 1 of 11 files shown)


views/clients/show.erb
```diff
@@ -1,53 +1,63 @@
...
+            <% @keys = resource_hash[0].values[0].keys - ['id'] %>
+            <tr>
+              <th>Name</th>
+              <% @keys.each do |k| %>
+                <% if k == 'DBSize' %>
+                  <th>Database Size</th>
+                <% elsif k == 'VCPUCount' %>
+                  <th>CPU Count</th>
+                <% else %>
+                  <th><%= k.underscore.humanize.titleize %></th>
```
ramereth: I ran into the following [error](https://github.com/osuosl/iam/files/1062633/NoMethodError.at._clients_3.pdf) using imported production data.

ramereth:
@hsolorza looking better! The only thing I see wrong is the DB sizes should be totaled up to ACTUAL sizes on the regular view. Not rounded up to 1.

ramereth:
@hsolorza actually, the same rule applies for everything in the non-billing view. It shouldn't be doubling anything if its drbd for example. That's only needed for billing.

ramereth:
@hsolorza the values look correct now thanks! Can you also please add units?

ramereth:
@hsolorza the CPU should actually have no units

Issue #274, opened 2017-06-13
VM resources should only be counted when vm is in an up or running state


ramereth:
VMs can be allocated but not running. When they are not running they should not be counted towards billing. So I recommend we do the following:

1. We probably should still show the VM in the default display view, but we should add a column which shows the state as up or down.
2. For the billing view, we should add a column which states how many days of data is being shown for each resource.

For the second one, say we have a VM that was running for 15 days but I'm looking at report for 30 days. It should show that the VM was using those resources for 15 days while the other VMs show 30 days. This allows us to quickly calculate prorated fees.

Now the question I have is, how is the value calculated as it's changed over time? i.e. What if a VM had used 10G of disk used for 10 days and then 20G for 20 days. How would it be shown in the billing view? Is it an average?

PR #276, opened 2017-06-20
WIP make sure we store booleans

Kennric:
fixes issue #274

## Changes in this PR.

- [X] Store true or false instead of coercing "true" and "false" into boolean db field

## Testing this PR.

1. docker-compose up test
2. run on staging

### Expected Output.

Data collected should include active and inactive nodes with the correct boolean true or false, instead of all "true"

```
$ rake spec
[...]
```

@osuosl/dev

### DIFF
--- a/lib/BasePlugin/plugin.rb
+++ b/lib/BasePlugin/plugin.rb
@@ -84,6 +84,10 @@ class BasePlugin
     NodeResource.find_or_create(name: name, type: type, project_id: project,
                                 cluster: cluster)[:id]
   end
+
+  def true?(obj)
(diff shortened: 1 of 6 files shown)


alxngyn:
++

do we need to run a script to change existing data?

Kennric:
@alxngyn no, I think existing data is correct - postgres seems to do the right thing, and will coerce "true" to a boolean true, and "false" to boolean false, but it's not clear what it does when you try to save a non-truthy/falsey string to it. This is just a little extra precaution.

hsolorza:
+1

Issue #279, opened 2017-06-21
active node plugin


Kennric:
Currently we store the active state of a ganeti node in every measurement plugin. We should instead make a plugin that stores the active state of a node as a measurement. 

When this is done, we can simplify the project_data method and client views significantly, while still displaying the number of days the node was active over the given time range.

PR #277, opened 2017-06-20
WIP Kennric/better queries

Kennric:
addresses issue #263

## Changes in this PR.

- [X] Query all measurements for all resources at once for a project
- [X] Update util and route methods to use new query

## Testing this PR.

1. docker-compose up test
2. docker-compose up dev
3. view localhost:4567/clients/*

### Expected Output.

A summary of client data, no timeouts.

```
$ rake spec
[...]
```

@osuosl/devs

PR #280, opened 2017-06-21
add a plugin to measure node state

Kennric:
fixes issue #279

## Changes in this PR.

- [X] Add NodeActive plugin

## Testing this PR.

1. docker-compose up test
2. NodeActive tests are run and pass
3. deploy to staging
4. observer active state measurements being stored

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/plugins/NodeActive/plugin.rb
+++ b/plugins/NodeActive/plugin.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+require 'logging'
+require 'sequel'
+require_relative '../../lib/BasePlugin/plugin.rb'
+require_relative '../../environment.rb'
+require_relative '../../models.rb'
+require_relative '../../logging/logs'
+
+# Disk Sizes data plugin
+class NodeActive < BasePlugin
+  def initialize
+    name = 'NodeActive'
+    resource_name = 'node'
+    units = 'boolean'
+    table = :node_active_measurements
+    db_column = :node_active_ver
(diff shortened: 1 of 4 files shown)


Kennric:
When this is merged, the code in #277 can be simplified, since we won't need to infer activity from other measurements

plugins/NodeActive/plugin.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+require 'logging'
+require 'sequel'
+require_relative '../../lib/BasePlugin/plugin.rb'
+require_relative '../../environment.rb'
+require_relative '../../models.rb'
+require_relative '../../logging/logs'
+
+# Disk Sizes data plugin
```
hsolorza: Should be Node Active

plugins/NodeActive/plugin.rb
```diff
@@ -0,0 +1,51 @@
...
+      MyLog.log.warn "NodeActive: No node state information for #{fqdn}"
+      raise "NodeActive: No active information for #{fqdn}"
+    end
+
+    # check if node resource exist, otherwise set it to default
+    node_resource = @database[:node_resources].where(name: fqdn).get(:id)
+    node_resource = default_node(fqdn, node_info['type'],
+                                 node_info['cluster']) unless node_resource
+
+    # Insert data into disk_size_measurements table
```
hsolorza: Should be node_active_measurements

alxngyn:
++

Issue #251, opened 2017-04-19
SKU - resource - project association


Kennric:
Currently node and db resources have a many-to-one relation to projects (Project has many *_resources). To associate a specific SKU with the specific Project-resource, we'll need an intermediate join table between project and and each resource type which stores the project_id, *_resource_id and a sku_id.

When adding a resource to a project, the form must include a selector for all available SKUs to add a sku to the project/resource association.

Probably the best way to accomplish this with the Sequel ORM is to create a model for the join table (a ProjectNodeResource model, for instance) which itself has a many-to-many association with SKUs

alxngyn:
Model // relation
- [ ] Test, 3
- [ ] Code, 5

Crud
- [ ] Test, 3
- [ ] Code, 3

PR #284, opened 2017-06-26
Model association crud

hsolorza:
fixes issue #251 

## Changes in this PR.

- [X] CRUD
- [X] Minor fixes to views
- [x] Tests

## Testing this PR.

1. Create a node/db
2. Assign it a SKU
3. Are able to delete either the SKU, node, or db without crash

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -22,9 +22,9 @@ class Iam < Sinatra::Base
   # to make sure every record is collected into the hierarchy
   default_client = Client.find_or_create(name: 'default',
                                          description: 'The default client')
-  default_project = Project.find_or_create(name: 'default',
-                                           client_id: default_client.id,
-                                           description: 'The default project')
+  Project.find_or_create(name: 'default',
+                         client_id: default_client.id,
+                         description: 'The default project')
 
   register Sinatra::MainRoutes
   register Sinatra::ClientRoutes
--- a/app.rb
+++ b/app.rb
@@ -22,9 +22,15 @@ class Iam < Sinatra::Base
   # to make sure every record is collected into the hierarchy
   default_client = Client.find_or_create(name: 'default',
                                          description: 'The default client')
-  Project.find_or_create(name: 'default',
-                         client_id: default_client.id,
-                         description: 'The default project')
+  default_project = Project.find_or_create(name: 'default',
+                                           client_id: default_client.id,
+                                           description: 'The default project')
+  default_sku = Sku.find_or_create(name: 'default',
+                                   description: 'The default SKU')
+  NodeResourcesProject.find_or_create(project_id: default_project.id,
+                                      sku_id: default_sku.id)
+  DbResourcesProject.find_or_create(project_id: default_project.id,
+                                    sku_id: default_sku.id)
 
   register Sinatra::MainRoutes
   register Sinatra::ClientRoutes
--- a/app.rb
+++ b/app.rb
@@ -25,12 +25,6 @@ class Iam < Sinatra::Base
   default_project = Project.find_or_create(name: 'default',
                                            client_id: default_client.id,
                                            description: 'The default project')
-  default_sku = Sku.find_or_create(name: 'default',
-                                   description: 'The default SKU')
-  NodeResourcesProject.find_or_create(project_id: default_project.id,
-                                      sku_id: default_sku.id)
-  DbResourcesProject.find_or_create(project_id: default_project.id,
-                                    sku_id: default_sku.id)
 
   register Sinatra::MainRoutes
   register Sinatra::ClientRoutes
--- a/models.rb
+++ b/models.rb
@@ -18,10 +18,8 @@ class Client < Sequel::Model
     errors.add(:name, 'cannot be empty') if !name || name.empty?
   end
 
-  def remove(default_project)
-    unless projects.empty?
-      projects.each { |project| project.reassign_resources default_project }
-    end
+  def remove
+    projects.each(&:reassign_resources) unless projects.empty?
     delete
   end
 end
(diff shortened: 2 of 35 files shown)


app.rb
```diff
@@ -22,9 +22,15 @@ class Iam < Sinatra::Base
...
                                          description: 'The default client')
-  Project.find_or_create(name: 'default',
-                         client_id: default_client.id,
-                         description: 'The default project')
+  default_project = Project.find_or_create(name: 'default',
+                                           client_id: default_client.id,
+                                           description: 'The default project')
+  default_sku = Sku.find_or_create(name: 'default',
+                                   description: 'The default SKU')
+  NodeResourcesProject.find_or_create(project_id: default_project.id,
```
Kennric: I don' know if this is necessary - first, these don't associate an actual node or db, so I am not sure what function they serve, and since we don't have a default node or default db, I don't know that we need any default associations created at all, just the project and client objects.

models.rb
```diff
@@ -18,8 +18,10 @@ def validate
     errors.add(:name, 'cannot be empty') if !name || name.empty?
   end
 
-  def remove
-    projects.each(&:reassign_resources) unless projects.empty?
+  def remove(default_project)
```
Kennric: We can always look up the default project by name, so I am not sure we need this change - in reassign_resources, just ```Project.find(name: 'default')`` since I think it will only ever be used to reassign to the system default project, which should always have the name 'default'

models.rb
```diff
@@ -114,6 +125,11 @@ def validate
     super
     errors.add(:name, 'cannot be empty') if !name || name.empty?
   end
+
+  def reassign_resources
+    DbResourcesProject.where(db_resource_id: id).delete
+    update(active: false)
```
Kennric: This caused me some confusion, since it isn't really reassigning resources, but just doing some extra things on delete. It should be renamed of the logic moved into the ``delete`` method for this model

routes/clients.rb
```diff
@@ -130,7 +130,8 @@ def self.registered(app)
       app.delete '/clients/:id/?' do
         # delete a client
         client = Client[id: params[:id]]
-        client.remove unless client.name == 'default'
+        default_project = Project.find(name: 'default').id
+        client.remove default_project unless client.name == 'default'
```
Kennric: Don't forget to remove the argument if you remove that parameter from the method as suggested

routes/database.rb
```diff
@@ -101,25 +109,30 @@ def self.registered(app)
...
+                          sku_id: params[:sku_id] || project_db.sku_id)
 
         redirect "/db/#{params[:id]}"
       end
 
       app.delete '/dbs/:id/?' do
         # delete a database
         db = DbResource[id: params[:id]]
-        db.update(active: false)
+        # disassociate this db sku's and deactivate the db
```
Kennric: I think maybe reassign_resources is not the right method name for this - reassign_resources is really just for when we delete projects and clients, so their resources get reassigned back to default. This is really a delete method that also removes the project and sku associations, which is a little different and could probably just be put in the ``delete`` method of the model. This may just be a naming issue, but it confused me when reviewing the reassign_resources methods in the models.

models.rb
```diff
@@ -85,6 +91,11 @@ def validate
     super
     errors.add(:name, 'cannot be empty') if !name || name.empty?
   end
+
+  def reassign_resources
```
Kennric: See comment on same method in NodeResource

routes/nodes.rb
```diff
@@ -117,25 +126,29 @@ def self.registered(app)
...
+                            sku_id: params[:sku_id] || project_node.sku_id)
         redirect "/node/#{params[:id]}"
       end
 
       app.delete '/nodes/:id/?' do
         # delete a node
         node = NodeResource[id: params[:id]]
-        node.update(active: false)
+        # disassociate this nodes sku's and deactivate the node
+        node.reassign_resources
```
Kennric: See comment on DB delete route

routes/projects.rb
```diff
@@ -93,9 +93,10 @@ def self.registered(app)
       app.delete '/projects/:id/?' do
         # delete a project
         project = Project[id: params[:id]]
+        default_id = Project.find(name: 'default').id
```
Kennric: Again, just do this in the reassign_resources method

routes/skus.rb
```diff
@@ -86,7 +86,7 @@ def self.registered(app)
       app.delete '/skus/:id/?' do
         # delete a sku
         sku = Sku[id: params[:id]]
-        sku.delete unless sku.nil?
+        sku.reassign_resources unless sku.name == 'default'
```
Kennric: As noted above, add the reassign to the delete method, and instead of just not doing anything for the default sku, it would be nice to send a message explaining that the default sku cannot be deleted.

Also, it is possible we don't need a default sku at all - resources can just not have one, right? It might be good to have a default one as a placeholder so no logic gets confused, but a nil in the sku field should not be hard to handle. If a default sku makes view logic simpler, keep it, otherwise lets not have one.

Kennric (changes requested):

app.rb
hsolorza: So, delete the default sku as well?

alxngyn (commented):
- deactivating a SKU does not update the related nodes attributes
that's the only bug i could find.

alxngyn (approved):
You fixed my comment
++

baileysingleton (approved):
Works as specified, a little confusing with the SKU edit stuff on the database, but when the edit PR Is merged I think it will all work out nicely.

++

PR #231, opened 2017-01-21
switch to bs3, fix header, and modify index

alxngyn:
fixes issue #230 #218 

## Changes in this PR.

- [X] Add something to the index page
- [X] Move to a stable version of bootstrap
- [x] Add statistics to index
- [x] Add pictures/color?
- [x] Add description details

## Testing this PR.

1. Do docker.
2. Do look at index `localhost:4567/`.
3. See view.

### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

PR #286, opened 2017-07-10
Updating Dates on testdata import

DevonCash:
**fixes issue #231**

## Changes in this PR.

- [x] Collects the latest date in records, calculates offset
- [X] Updates all dates with offset upon import
- [X] Stopped slurping data files, reading line-by-line

## Testing this PR.

1. Get some sample data
2. Run 'rake import_data'
3. View test data with recent updates

@osuosl/devs

### DIFF
--- a/lib/datasampler.rb
+++ b/lib/datasampler.rb
@@ -6,10 +6,12 @@ require_relative './util.rb'
 
 def iterate_lines(file_path)
   return unless File.file?(file_path)
-  File.foreach(file_path, '}') do |line|
+  lines = 0
+  File.foreach(file_path, '}').with_index do |line|
     line = line[/{.+}/] # Remove obj delimeters
     next if line.nil?
-    yield(line)
+    lines += 1
+    yield(lines, line)
   end
 end
 
@@ -49,8 +51,13 @@ class DataImporter
   # rubocop:disable MethodLength
   def date_offset
     latest_date = nil
+    @count = {}
     Dir.foreach(@directory) do |file|
-      iterate_lines(@directory + file) do |line|
+      next if file=='.' || file=='..'
+      print "\n  #{file}..."
+      @count[file] = 0
+      iterate_lines(@directory + file) do |idx, line|
+        @count[file] += 1
         JSON.parse(line).each do |_key, value|
           next unless value.is_a? String
           begin
@@ -62,6 +69,7 @@ class DataImporter
         end
       end
     end
+    print "\n"
     @date_offset = DateTime.now - latest_date
   end
 
@@ -69,9 +77,11 @@ class DataImporter
   def import_clients
     clients_filename = @directory + 'clients.json'
     Client.unrestrict_primary_key
-    iterate_lines(clients_filename) do |line|
+    iterate_lines(clients_filename) do |idx, line|
+      print "Importing Clients #{idx}/#{@count['clients.json']}\r"
       Client.create(JSON.parse(line))
     end
+    print "\n"
     Client.restrict_primary_key
   end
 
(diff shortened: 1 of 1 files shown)


lib/datasampler.rb
```diff
@@ -25,70 +41,116 @@ def delete_data
...
   def import_clients
-    # get clients from file, import to Client model
     clients_filename = @directory + 'clients.json'
-    clients_json = File.open(clients_filename, &:readline)
-    clients = Client.array_from_json(clients_json,
-                                     fields: Client.columns.map(&:to_s))
-
-    clients.each(&:save)
+    Client.unrestrict_primary_key
+    iterate_lines(clients_filename) do |line|
```
Kennric: I am still not sure why you are doing this line-by-line.

lib/datasampler.rb
```diff
@@ -25,70 +41,116 @@ def delete_data
...
-        Iam.settings.DB[table.to_sym].multi_insert(measurement_data)
+      next unless File.exist? filename
+      puts "Importing resource #{resource_name} from #{filename}"
+
+      model = Object.const_get((resource_name + 'Resource').camelcase(:upper))
+      model.unrestrict_primary_key
+      iterate_lines(filename) do |line|
+        resource = JSON.parse(line)
+        resource.each do |key, value|
+          value = update_date(value, @date_offset)
```
Kennric: We probably don't need to update the date on resources - it shouldn't hurt anything, but we only ever care about the measurement time

Kennric (commented):
This all looks pretty good, though I am not sure about the line-by-line processing, I am not sure why that's easier than importing a single json array and using ruby's hash/array methods to update dates, etc. As long as it works and is not enormously slower, I don't care that much, the code looks pretty solid. 

I'll dump some data with your modified export methods and we can test some more.

lib/datasampler.rb
```diff
@@ -166,14 +228,18 @@ def export_data(client_list: ['all'], days: 60, anon: true)
...
+    end
 
     all_projects = Project.where(id: project_ids).naked.all
     anonymize_projects(all_projects) if anon
 
     # write the projects to a file in json fromat
     filename = @directory + 'projects.json'
-    File.open(filename, 'w') { |file| file.write(all_projects.to_json) }
+    File.open(filename, 'w') do |file|
+      file.write(JSON.generate(all_projects))
```
panntomime: From what I've seen, that would only be an issue if we were directly importing the json into sequel. Since we're processing it first, reading it into ruby, then pushing into sequel, we're passing a ruby hash in rather than a string

lib/datasampler.rb
panntomime: I'm doing it line by line in an effort to speed it up a little. Beforehand, we were reading the entire file into memory first, then iterating over it. This way, much less RAM is used. I optimized this further in a later commit by aggregating the result of the change in an array, then using Sequel's import function to reduce the number of transactions used

Kennric:
It looks like you have some Rubocop errors - be sure to run ``rvm all do rubocop`` before pushing up (if you don't have rvm set up locally, do that, and ``rvm all do bundle install`` to make sure you have the right version of rubocop)

hsolorza (changes requested):
I'm getting a NameError when viewing a client's page
screenshot from 2017-07-26 12-03-11

panntomime:
> I'm getting a NameError when viewing a client's page
This branch wasn't merged with develop yet, so didn't have the NodeActive plugin at all. Merged in order to fix this view

image

hsolorza (approved):

Issue #288, opened 2017-07-24
Edit form does not allow for new project


hsolorza:
When editing  a node or database, the form for changing the project does not actually update the new assigned project.

PR #289, opened 2017-07-24
Hsolorza/edit changes

hsolorza:
fixes issue #288 

## Changes in this PR.

- [X] Creates dropdown for node/db project edit

## Testing this PR.

1. Create node/ db
2. Assign the node/db a project
3. Edit the node/db to have a different project
4. See the updated project assignment

NOTE: this PR only works correctly if the caching is deactivated
### Expected Output.

```
$ rake spec
[...]
```

@osuosl/devs

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -18,7 +18,6 @@ class Iam < Sinatra::Base
     cache_control :public, :must_revalidate, max_age: Iam.settings.cache_max_age
   end
 
-
   # initialize the app with some default clients, projects and nodeResources
   # to make sure every record is collected into the hierarchy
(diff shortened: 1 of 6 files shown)


alxngyn (approved):
looks good.
works as described

panntomime:
The form control looks good, but it doesn't look like the project is updating on the Node Detail page -- Is that expected behavior?

hsolorza:
@panntomime  I think that occurs because of the caching. If you disable it (in app.rb and config.rb) it should work.

baileysingleton (approved):
Everything works as specified, after I commented out the cache that is 📦 
++

